PR #1, opened 2019-01-28
First pass lexical model compiler

mcdurdin:


release/example/en.wordlist/source/model.ts
```diff
@@ -0,0 +1,11 @@
+import LexicalModelCompiler from "../../../../tools/index";
+
+(new LexicalModelCompiler).compile({
+  format: 'trie-1.0',
+  wordBreaking: {
+    allowedCharacters: { initials: 'abcdefghijklmnopqrstuvwxyz', medials: 'abcdefghijklmnopqrstuvwxyz', finals: 'abcdefghijklmnopqrstuvwxyz' },
```
eddieantonio: Although I like how compact this notation is, JavaScript's strings may not work well, because there's no explicit way to denote full characters that are greater than one code point (e.g., <X̱> in Northern Haida, two code points in NFC). Also, characters that are outside the BMP will be a pain, and error-prone.

release/example/en.custom/source/wordbreak.ts
```diff
@@ -0,0 +1,18 @@
...
+    var match, indexes = [];
+
+    while (match = regex.exec(string)) {
+      indexes.push(match.index);
+    }
+
+    return indexes;
+  }
+
+  return indexesOf(text);
```
eddieantonio: Do you feel the indices in the string will be more useful than returning the actual words as strings? If word breaker functions return both the start and the end index of a word, then we can have it both ways: default is to return indices, and a simple function call can convert indices into words as strings.

resources/util.sh
```diff
@@ -0,0 +1,119 @@
+#!/bin/bash
+
+#
+# Define terminal colours
+#
+
+if [ -t 2 ]; then
```
eddieantonio: I had to look up what `[ -t 2]`. Could you please provide a comment explain why this is here? I reckon it's to define ANSI colours ONLY when `stderr` is outputting to the terminal (as opposed to redirected to a file).

release/example/en.wordlist/source/model.ts
mcdurdin: So could we support string -- for simple situations or array of strings for others? e.g. `['a','b',...,'X̱']`?

I am unsure on how this is going to be used -- I would happy to have the source format support both but compiled into array of strings for simplicity of consumption.

release/example/en.custom/source/wordbreak.ts
mcdurdin: Good questions. Am happy to adjust this per the requirements of the wordbreaking when we get to it. I've defaulted to the most basic approach possible now of just marking the characters which are wordbreakers -- but I know that is naive. There are plenty of other wordbreak algorithms that we should research as well (yay!)

resources/util.sh
mcdurdin: Sure :)

tools/index.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    //
+    // Build the compiled lexical model
+    //
+    
+    let sources: string[] = o.sources.map(function(source) { 
+      return fs.readFileSync(path.join(sourcePath, source), 'utf8'); 
+    });
+
+    let oc: LexicalModelCompiled = {id:model_info.id, format:o.format, wordBreaking:o.wordBreaking};
```
jahorton: `format` and `wordBreaking` are not listed as properties in tools/lexical-model.ts.

release/example/en.wordlist/source/model.ts
eddieantonio: I like this idea!

tools/index.ts
mcdurdin: `interface LexicalModelCompiled extends LexicalModel` :)

tools/index.ts
```diff
@@ -0,0 +1,157 @@
...
+      this.logError('Unable to find .model_info file in parent folder');
+      return false;
+    }
+    
+    let model_info: ModelInfoFile = JSON.parse(fs.readFileSync('../'+model_info_file, 'utf8'));
+
+    //
+    // Filename expectations
+    //
+    const kpsFileName = '../source/'+model_info.id+'.model.kps';
```
eddieantonio: Suggestion: String templates may be cleaner here:

```typescript
const kpsFileName = `../source/${model_info.id}.model.kps`;
```

tools/kmp-json-file.ts
```diff
@@ -0,0 +1,79 @@
+interface KmpJsonFile {
```
eddieantonio: Where will the official documentation (i.e., meaning of each field) for the JSON file exist?

tools/lexical-model.ts
```diff
@@ -0,0 +1,35 @@
+
+interface LexicalModel {
+  readonly format: 'trie-1.0'|'fst-foma-1.0'|'custom-1.0',
+  readonly wordBreaking?: {
```
eddieantonio: Note that a custom model may provide its own word breaking.

tools/lexical-model.ts
```diff
@@ -0,0 +1,35 @@
...
+    allowedCharacters?: { initials?: string, medials?: string, finals?: string } | string,
+    defaultBreakCharacter?: string
+    sources?: Array<string>;
+  },
+  //... metadata ...
+}
+
+interface LexicalModelPrediction {
```
eddieantonio: Are these interfaces to coorespond with the interfaces in the main repo?

See: https://github.com/keymanapp/keyman/blob/master/common/predictive-text/message.d.ts#L145-L224

tools/lexical-model.ts
```diff
@@ -0,0 +1,35 @@
...
+
+interface LexicalModelCompiledTrie extends LexicalModelCompiled {
+  trie: string;
+}
+
+interface LexicalModelCompiledFst extends LexicalModelCompiled {
+  fst: string;
+}
+
+interface LexicalModelCompiledCustom extends LexicalModelCompiled {
```
eddieantonio: This will most likely also contain `src: string`, but we'll cross that bridge when we get there :D

tools/model-info-file.ts
```diff
@@ -0,0 +1,33 @@
+interface ModelInfoFile {
```
eddieantonio: Where will the official documentation for this live?

eddieantonio:
My major concerns so far are: where the JSON documentation will exist; and whether the prediction interfaces will be duplicated and compatible with those in [keymanapp/keyman](https://github.com/keymanapp/keyman/blob/master/common/predictive-text/message.d.ts#L145-L224).

Aside from that, LGTM!

tools/kmp-json-file.ts
mcdurdin: I have an open PR on the help site (currently private, although shortly hoping to move it public) which documents these fields. See https://help.keyman.com/developer/11.0/reference/file-types/metadata (probably will be in 12.0 URL when that lands).

tools/lexical-model.ts
mcdurdin: Yes, this is really a stub for the real thing. I will update `LexicalModelPrediction` to correspond more cleanly with the `Transform` interface (but will not rely on the one in the main repo for now -- refactor can come once everything stabilises I think).

tools/model-info-file.ts
mcdurdin: It will live in the https://help.keyman.com/developer/cloud/ area, alongside the https://help.keyman.com/DEVELOPER/cloud/keyboard_info/1.0/ .keyboard_info file. Again, I have a PR open for this.

There is also a PR open against api.keyman.com/schemas for the corresponding JSON schema (and the updated kmp.json schema as well). Also hopefully to go public soon.

mcdurdin: * https://github.com/keymanapp/api.keyman.com/pull/1 covers the JSON schema and changes to package metadata format.
* https://github.com/keymanapp/help.keyman.com/pull/1 includes the documentation for the .model_info file (as of now, WIP).

These PRs are both #1 because we've just moved api and help sites to public source.

mcdurdin:
Okay, I've addressed a bunch of bits and pieces with the compiler and documentation, and tried to cover the review comments. At this point, there are still a bunch of unfinished bits and pieces but I think that's fine, because this still gives us a base for generating .kmp files for use in the target applications, and for establishing CI.

resources/compile.sh
```diff
@@ -135,7 +135,7 @@ function build_release_model {
 
   pushd build
   mkdir obj
-  ../../../../node_modules/.bin/tsc --outDir ./obj ../source/model.ts 
+  ../../../../node_modules/.bin/tsc --module commonjs --target es6 --outDir ./obj ../source/model.ts
```
eddieantonio: 1. would it be cleaner to use `npx tsc` instead of `../../../../node_modules/.bin/tsc`>
2. Is it worth extracting the compiler options to an explicit `tsconfig.json`?

mcdurdin: 1. Updated to use `npx`. Note, on Windows, this requires Node.js version 10.0 or later.

2. Am I right in thinking this would require an explicit `tsconfig.json` for each model? If so for now I'll leave it in the script.

PR #2, opened 2019-02-27
[meta] continuous integration script (clone of keyboards script)

mcdurdin:


ci.sh
```diff
@@ -0,0 +1,223 @@
...
+
+  if [[ $DO_ZIP_ONLY == true ]]; then
+    if [ ! -d "$CI_CACHE/data" ]; then
+      mkdir "$CI_CACHE/data"
+    fi
+    if [ ! -d "$CI_CACHE/upload" ]; then
+      mkdir "$CI_CACHE/upload"
+    fi
+    zip_model_info
+    rsync_to_downloads_keyman_com "$CI_CACHE/data/" model-data/
```
darcywong00: Why is this rsync inside the `$DO_ZIP_ONLY` block?

mcdurdin: `$DO_ZIP_ONLY` might be a misnomer -- it means upload only the model metadata.

mcdurdin: This is also only really used for testing

resources/rsync-tools.sh
```diff
@@ -0,0 +1,32 @@
...
+  else
+    ignore=
+  fi
+  
+  #rsync_args = (
+  #  '-vtrp',                                  # verbose, preserve times, recurse, permissions
+  #  '--chmod=Dug=rwx,Do=rx,Fug=rw,Fo=r',      # map Windows security to host security
+  #  '--stats',                                # show statistics for log
+  #  "--rsync-path=$REMOTE_RSYNC_PATH",        # path on remote server
+  #  "--rsh=ssh",                              # use ssh
```
darcywong00: can add
```
# "--ignore-existing",                   # .....
```

PR #3, opened 2019-03-05
Reconfigures existing model compilation structure.

jahorton:
This PR is the start of work on compiling model files to a form (mostly) friendly to the LMLayer's current state.  (The LMLayer doesn't yet implement a word-breaking model.)

Still needed:
* LMLayerWorker code to handle word-breaking and word-breaker loading (on keymanapp/keyman repo)
  * See https://github.com/keymanapp/keyman/pull/1651.
* Further refinement of compilation.  (Can be done in later PRs.)


tools/index.ts
```diff
@@ -86,46 +86,54 @@ export default class LexicalModelCompiler {
...
         (oc as LexicalModelCompiledFst).fst = Buffer.from(sources.join('')).toString('base64');
         this.logError('Unimplemented model format '+o.format);
         return false;
       case "trie-1.0":
+        func += `var model = {};\n`;
         // TODO: compile the trie
-        (oc as LexicalModelCompiledTrie).trie = sources.join(' ');
-        func += funcPrefix + JSON.stringify(oc) + funcSuffix;
+        func += `model.backingData = ${JSON.stringify(sources.join(' '))};\n`;
+        func += `LMLayerWorker.loadModel(new models.WordListModel({}, model.backingData));\n`;
```
eddieantonio: 👍

tools/lexical-model.ts
```diff
@@ -5,6 +5,7 @@ interface LexicalModel {
     allowedCharacters?: { initials?: string, medials?: string, finals?: string } | string,
     defaultBreakCharacter?: string
     sources?: Array<string>;
+    root?: string
```
eddieantonio: Throughout this PR, I'm unclear exactly what `root` is, and why we're accessing `root` everywhere. Could you please document this parameter?

jahorton:
Sure.  I might should figure out a better name for it - maybe `rootClass` - but it's the name of the TS class that implements the `WorkerInternalModel` or `WorkerInternalWordBreaker` (in reference to https://github.com/keymanapp/keyman/pull/1651) that must be supplied to `LMLayerWorker.loadModel` / `LMLayerWorker.loadWordBreaker` respectively.

It'd be nice to not need the parameter, but I see no simple, reliable way for the compiler to 'intuit' the correct type from within the source at this time, hence the need for said class name to be supplied as part of model.ts's configuration specs.

includes/LMLayer.d.ts
```diff
@@ -0,0 +1,72 @@
...
+    activateModel(model: string): Promise<Configuration>;
+
+    /**
+     * Prepares the LMLayer for reinitialization with a different model/capability set.
+     */
+    deactivateModel(): void;
+
+    predict(transform: Transform, context: Context): Promise<Suggestion[]>;
+
+    // TODO: asynchronous close() method.
```
eddieantonio:
```suggestion
    // TODO: asynchronous close() method.
```

includes/LMLayer.d.ts
```diff
@@ -0,0 +1,72 @@
...
+    //       Worker code must recognize message and call self.close().
```
eddieantonio:
```suggestion
```

includes/LMLayer.d.ts
```diff
@@ -0,0 +1,72 @@
...
+
```
eddieantonio:
```suggestion
```

release/example/en.custom/source/model.ts
```diff
@@ -3,8 +3,10 @@ import LexicalModelCompiler from "../../../../tools/index";
...
   format: 'custom-1.0',
   wordBreaking: {
-    sources: ['wordbreak.ts']
+    sources: ['wordbreak.ts'],
+    rootClass: 'ExampleWordBreaker'
   },
   //... metadata ...
-  sources: ['predict.ts']
+  sources: ['predict.ts'],
+  rootClass: 'ExampleCustom'
```
eddieantonio: The name `rootClass` makes more sense than `root`! 👍

release/example/en.custom/source/wordbreak.ts
```diff
@@ -1,18 +1,21 @@
...
+      while (match = regex.exec(string)) {
+        tokens.push(match[0]);
+      }
+  
+      return tokens;
     }
-
-    return indexes;
+  
+    return tokensOf(text);
```
eddieantonio: There are a few ways that we could design this API. One is to return the start indices (as before), and another is to return the tokens themselves. Another way is to return both start and end indicies, AND the token token text. One of these attributes could even be an accessor rather than a "real" attribute of the class, e.g.,

```typescript
class WordBreak {
  start: number;
  end: number;
  private source: string;
  get token() {
     return this.source.substr(this.start, this.end);
  }
}
```

This does **not** need to be addressed in this PR, but I'm just putting it out there. I still don't know what interface is the most convenient for implementors.

 Does not need to be addressed in

jahorton: Yeah, I'm not exactly sure on which way to go.  Tokenization is probably the easiest to consume for lexical models - hence why I chose it - though it might get in the way when the time comes to build a new `Transform` to apply for a suggestion, since it doesn't supply indices.

Worst case, there's always the "why not both?" option.  But yeah, that can be addressed in followup work; we've yet to implement a default word-breaker, after all.

darcywong00:
Just checking if `en.custom` version needs to increment to 1.0.1

mcdurdin:
Good point @darcywong00. The distributed model files will not be updated if the version number is not incremented. We can do a cleanup pre-stable in order to avoid having irrelevant old versions of the example models on the downloads server.

release/example/en.custom/source/predict.ts
```diff
@@ -1,10 +1,27 @@
...
+
+  configure(capabilities: Capabilities) {
+    return this.configuration = {
+      leftContextCodeUnits: 64 < capabilities.maxLeftContextCodeUnits ? 64 : capabilities.maxLeftContextCodeUnits,
+      rightContextCodeUnits: 0
+    }
   }
-});
+
+  // TODO:  Needs to reference a .d.ts file for type definitions!
```
mcdurdin: Has this TODO been addressed already?

release/example/en.custom/source/wordbreak.ts
```diff
@@ -1,18 +1,21 @@
...
-com.keyman.lexicalModel.registerWordBreaker('example.en.custom', function(text: string): Array<number> {
-  let regex = new RegExp(/[!a-zA-Z]/g);
-
-  let indexesOf = function(string) {
-    var match, indexes = [];
-
-    while (match = regex.exec(string)) {
-      indexes.push(match.index);
+class ExampleWordBreaker {
+  // TODO:  Needs to reference a .d.ts file for type definitions!
```
mcdurdin: Has this TODO been addressed already?

mcdurdin:
Note: before merge, I want to enable pull request builds.

mcdurdin:
Okay, Pull Request builds are running correctly (now have permissions sorted so they can report back on the commit/PR).

tools/index.ts
```diff
@@ -86,46 +86,54 @@ export default class LexicalModelCompiler {
...
-
     //
     // Emit the model as code and data
     //
 
     switch(o.format) {
       case "custom-1.0":
-        func += funcPrefix + JSON.stringify(oc) + funcSuffix + '\n' + this.transpileSources(sources).join('\n');
+        func += funcPrefix + funcSuffix + '\n' + this.transpileSources(sources).join('\n');
+        // JSON.stringify(oc) gives the base metadata
```
eddieantonio: Comment appears to no longer be relevant.

```suggestion
```

PR #11, opened 2019-03-25
Infer model ID from .model_info filename

### LINKED ISSUE #10: Currently the compiler requires `id` to be present in the source `.model_info`
Currently the compiler does require `id` to be present in the source `.model_info` which is technically a bug; we should instead pull that from the folder name. I will open an issue.

This is a problem because:

1. The [spec](https://help.keyman.com/developer/cloud/model_info/1.0/) says that `id` is not required.
2. It's WETter than it needs to be.

_Originally posted by @mcdurdin in https://github.com/keymanapp/lexical-models/pull/7#issuecomment-476359117_

eddieantonio:
Fixes #10


eddieantonio:
@mcdurdin, you'll be pleased to know that TeamCity actually built this PR! Let me know what you think!

eddieantonio:
> Note that we can still remove some duplicated data in the .model_info files I think by putting that same data into the .kps file?

Sounds good! I'll leave that for a different PR, though.

PR #13, opened 2019-03-27
Fix ID to crk.wordlist_wahkohtowin

darcywong00:
Fix the typo in the model ID which was preventing the mobile apps from installing the model.


darcywong00:
This is merged. @mcdurdin will investigate 
1. why the build server kept the model at version 1.0.0 (should be 1.0.1)
2. updating the compiler to flag when the model ID doesn't match

mcdurdin:
1. Looks to be fixed now.
2. Done in #18

PR #14, opened 2019-03-27
[crk] Bump model version in .model_info

darcywong00:
#13 fixed the model ID, but I forgot to bump the model version in the .model_info file

### DIFF
--- a/release/example/crk.wordlist_wahkohtowin/example.crk.wordlist_wahkohtowin.model_info
+++ b/release/example/crk.wordlist_wahkohtowin/example.crk.wordlist_wahkohtowin.model_info
@@ -1,7 +1,7 @@
 {
     "name": "Example Plains Cree Wordlist Model",
     "license": "mit",
-    "version": "1.0.0",
+    "version": "1.0.1",
     "languages": [
         "crk"
     ],


mcdurdin:
So. A few things:

1. The merge of the model info should warn if there are mismatching version numbers, and ...
2. The version number shouldn't be in the .model_info, and ...
3. The canonical data is always in the source files, not the .model_info, because ...
4. Lexical models can be built, stored and distributed outside the `lexical-models` repository and CI chain. Those lexical models won't have a .model_info file, because that is used only for the CI deployment. It is important we don't end up building too many dependencies on the repository structure by accident because it is likely, if we are successful, that this repo will get very large. The intent is for models in this repo to be of high release-level quality and we generally want to discourage experimental models here, because models here can be automatically installed by Keyman. This leads me to conclude that ...
5. The compiler needs to be moved out of the lexical-models repo and into the keyman repo, and the .model_info management needs to be clearly delineated. Now, the compiler is part of the Keyman Developer toolchain. So I think what we need to do is update the build process to pull the latest stable (or `$tier`) version of the compiler from downloads.keyman.com (so that's something that needs to be added to the Keyman Developer CI). We *could* register an NPM package but I don't think we are quite ready to do that; happy to defer to wiser heads though. For consistency, then ...
6. This same process needs to be done with the keyboards repo -- so we don't include the kmcomp compiler in the repo, rather pull it at first build (or at appropriate times) from downloads.keyman.com.

I will shortly turn some of this into new issue(s) on the relevant repos. Any thoughts on this much appreciated before I go ahead and do that.

darcywong00:
Closing because the model versions are revamped in #17

eddieantonio:
> * *The compiler needs to be moved out of the lexical-models repo and into the keyman repo*, and the .model_info management needs to be clearly delineated. Now, the compiler is part of the Keyman Developer toolchain. So I think what we need to do is update the build process to pull the latest stable (or `$tier`) version of the compiler from downloads.keyman.com (so that's something that needs to be added to the Keyman Developer CI). We _could_ register an NPM package but I don't think we are quite ready to do that; happy to defer to wiser heads though. For consistency, then ...

This sounds reasonable to me! I like the idea of an `npm` package: something like `@keyman/lexical-model-compiler`. We can host it on `downloads.keyman.com` for now.

PR #16, opened 2019-03-28
Fixes .kmp bundling.

### LINKED ISSUE #15: Package compilation improperly bundles resources
While the individual resources for each model are properly built at present, the .model.kmp package file itself is not.  The issue:  both welcome.htm and the *.js file with compiled model code within the .model.kmp are instead **_relative paths_** to the individual build products, rather than copies of the actual files themselves.

Upon trying to actually utilize the .kmps for development work in our apps, we get something like the following:

image

As you might imagine, this rapidly generates errors and results in an unusable model.

jahorton:
Fixes #15.

The .kps entries for readme.htm were removed so that they didn't break .kmp compilation; the corresponding files do not yet exist in our examples.

### DIFF
--- a/release/example/crk.wordlist_wahkohtowin/source/example.crk.wordlist_wahkohtowin.model.kps
+++ b/release/example/crk.wordlist_wahkohtowin/source/example.crk.wordlist_wahkohtowin.model.kps
@@ -19,12 +19,6 @@
       <CopyLocation>0</CopyLocation>
       <FileType>.model.js</FileType>
     </File>
-    <File>
-      <Name>readme.html</Name>
-      <Description>Readme file</Description>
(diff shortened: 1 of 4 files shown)


jahorton:
@mcdurdin That is true.  The compiler did not check to see if readme.htm actually existed - instead, it created its own (useless) readme.htm within the .kmp based only on the path string.

mcdurdin:
So with this fix, I guess the build falls over if the file is missing?

jahorton:
I figured that would be a standard solution; that's what builds usually do, right?  Why define it as part of the package if you're not defining it?

mcdurdin:
Yes, that's what I want, just making sure :)

PR #18, opened 2019-03-29
Add some extra validation of source .model_info

mcdurdin:


tools/index.ts
```diff
@@ -177,21 +177,47 @@ export default class LexicalModelCompiler {
     // https://help.keyman.com/developer/cloud/model_info/1.0
     //
 
-    model_info.id = model_id; 
-    model_info.name = model_info.name || kmpJsonData.info.name.description;
-    model_info.authorName = model_info.authorName || kmpJsonData.info.author.description;
-    model_info.authorEmail = model_info.authorEmail || kmpJsonData.info.author.url;
+    function set_model_metadata(field, expected, warn?: boolean) {
```
eddieantonio:
```suggestion
    function set_model_metadata(field: string, expected: Any, warn?: boolean) {
```

tools/index.ts
```diff
@@ -177,21 +177,47 @@ export default class LexicalModelCompiler {
...
+      if(model_info[field] && model_info[field] !== expected) {
+        if(warn || typeof warn === 'undefined')
+          console.warn(`Warning: source ${modelInfoFileName} field ${field} value "${model_info[field]}" does not match "${expected}" found in source file metadata.`);
```
eddieantonio: 👍

tools/index.ts
```diff
@@ -177,21 +177,47 @@ export default class LexicalModelCompiler {
...
+    // Merge model info file -- some fields have "special" behaviours -- see below
+
+    set_model_metadata('id', model_id);
+    set_model_metadata('name', kmpJsonData.info.name.description);
+    set_model_metadata('authorName', kmpJsonData.info.author.description);
+
+    // we strip the mailto: from the .kps file for the .model_info
+    set_model_metadata('authorEmail', kmpJsonData.info.author.url.match(/^(mailto\:)?(.+)$/)[2], false);
+    
+    // TODO: languages is an array so more complicated to compare
```
eddieantonio: This map of maps is difficult to comprehend. Could you refactor it to make it simpler, or add a comment explain what its purpose is?

tools/index.ts
```diff
@@ -177,21 +177,47 @@ export default class LexicalModelCompiler {
...
-    model_info.lastModifiedDate = model_info.lastModifiedDate || (new Date).toISOString();
-    model_info.packageFilename = model_info.packageFilename || kmpFileName;
-    model_info.packageFileSize = fs.statSync(model_info.packageFilename).size; // Always overwrite with actual file size
-    model_info.jsFilename = model_info.jsFilename || modelFileName;
-    model_info.jsFileSize = fs.statSync(model_info.jsFilename).size; // Always overwrite with actual file size
-    model_info.packageIncludes = kmpJsonData.files.filter((e) => !!e.name.match(/.[ot]tf$/i)).length ? ['fonts'] : []; // Always overwrite source data
-    model_info.version = model_info.version || kmpJsonData.info.version.description;
-    model_info.minKeymanVersion = model_info.minKeymanVersion || minKeymanVersion;
+
+    set_model_metadata('lastModifiedDate', (new Date).toISOString());
```
eddieantonio: If this field is set in the .model_info, this line ensures it will practically never match expected! Should we state in the documentation (https://help.keyman.com/DEVELOPER/cloud/model_info/1.0/) that this field should not be explicitly specified? It seems to me like it will always produce a warning.

tools/index.ts
```diff
@@ -177,21 +177,47 @@ export default class LexicalModelCompiler {
...
+    // Always overwrite with actual file size
+    model_info.packageFileSize = fs.statSync(model_info.packageFilename).size; 
+
+    set_model_metadata('jsFilename', modelFileName);
+
+    // Always overwrite with actual file size
+    model_info.jsFileSize = fs.statSync(model_info.jsFilename).size;
+
+    // Always overwrite source data
+    model_info.packageIncludes = kmpJsonData.files.filter((e) => !!e.name.match(/.[ot]tf$/i)).length ? ['fonts'] : [];
```
eddieantonio: Will lexical models include fonts? Is that a now feature or a later feature?

tools/index.ts
mcdurdin: :) Yeah, this is really last reuploaded date, isn't it.

tools/index.ts
mcdurdin: Lexical models can include fonts, and this should be supported by the apps now.

tools/index.ts
mcdurdin: I'm not sure which map of maps you are referring to -- the languages property or something else? Can you elaborate?

mcdurdin: Ah sorry, I got it now -- your comment was on the `// comment ` line and I confused myself more than I needed to!

tools/index.ts
mcdurdin: See https://github.com/keymanapp/help.keyman.com/pull/25

mcdurdin:
Okay, I think I've addressed all your comments!

PR #23, opened 2019-04-23
Fix path separators in .kps files

darcywong00:
Fixes #3 so lexical models can be built on other platforms

I'm not updating model versions since the compiled artifacts should be the same (files are in a flat structure inside the .kmp)

### DIFF
--- a/release/example/crk.wordlist_wahkohtowin/source/example.crk.wordlist_wahkohtowin.model.kps
+++ b/release/example/crk.wordlist_wahkohtowin/source/example.crk.wordlist_wahkohtowin.model.kps
@@ -15,7 +15,7 @@
   </Info>
   <Files>
     <File>
-      <Name>../build/example.crk.wordlist_wahkohtowin.model.js</Name>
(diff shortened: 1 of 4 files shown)


release/example/crk.wordlist_wahkohtowin/source/example.crk.wordlist_wahkohtowin.model.kps
```diff
@@ -15,7 +15,7 @@
   </Info>
   <Files>
     <File>
-      <Name>..\build\example.crk.wordlist_wahkohtowin.model.js</Name>
```
mcdurdin: According to the .kps spec, these should be backslashes. This is not ideal, but it is what it is. The compiler needs to transform them instead.

darcywong00: alright, I added a regex for the slashes

eddieantonio:
@darcywong00 Does this fix #23, not #3?

darcywong00:
My bad. I've fixed the associated issue number.

PR #25, opened 2019-04-30
Add str-Latn to BCP 47 of SENĆOŦEN model.

eddieantonio:
This should help [fv_sencoten](https://help.keyman.com/keyboard/fv_sencoten/9.1/fv_sencoten.php)  consider this this an appropriate model.

### DIFF
--- a/release/example/str.trie_sencoten/example.str.trie_sencoten.model_info
+++ b/release/example/str.trie_sencoten/example.str.trie_sencoten.model_info
@@ -1,11 +1,4 @@
 {
-    "name": "Example SENĆOŦEN Trie Model",
     "license": "mit",
-    "languages": [
-        "str",
-        "str-Latn"
-    ],
-    "authorName": "Eddie Antonio Santos",
-    "authorEmail": "Eddie.Santos@nrc-cnrc.gc.ca",
     "description": "Example trie model for SENĆOŦEN"
 }
--- a/release/example/str.trie_sencoten/example.str.trie_sencoten.model_info
+++ b/release/example/str.trie_sencoten/example.str.trie_sencoten.model_info
@@ -2,7 +2,8 @@
     "name": "Example SENĆOŦEN Trie Model",
     "license": "mit",
     "languages": [
-        "str"
+        "str",
+        "str-Latn"
     ],
     "authorName": "Eddie Antonio Santos",
     "authorEmail": "Eddie.Santos@nrc-cnrc.gc.ca",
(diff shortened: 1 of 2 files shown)


eddieantonio:
@darcywong00 Done!

My, that's a might WET process! @mcdurdin What are the plans to streamline all of this juggling of `.model_info` and `.kps` consistency? I know there are a few checks in `tools/index.ts`, but it looks like there could be more. Also, we're intending that most of these files are generated by Keyman Developer, correct?

release/example/str.trie_sencoten/example.str.trie_sencoten.model_info
```diff
@@ -2,7 +2,8 @@
     "name": "Example SENĆOŦEN Trie Model",
     "license": "mit",
     "languages": [
-        "str"
+        "str",
+        "str-Latn"
```
mcdurdin: Languages will be sourced from the .kps if not included here. Nearly all fields in .model_info are sourced from .kps -- see https://help.keyman.com/developer/cloud/model_info/1.0/

The intent is that only metadata needed for distribution on the website is included in the source .model_info. The destination .model_info is built from that and from the .kps data. This should make it pretty DRY

eddieantonio: Ah! Somehow I didn't realize all of that from the table. 😖 Is the best course of action then to explicitly provide just the license and the description in the `.model_info` files?

mcdurdin:
Yes

eddieantonio:
Alright! The `.model_info` is a lot smaller now!

release/example/str.trie_sencoten/source/example.str.trie_sencoten.model.kps
```diff
@@ -27,7 +27,8 @@
       <ID>example.str.trie</ID>
       <Version>0.1.0</Version>
```
darcywong00: I think this version might need to bump to 0.1.1 too

mcdurdin: Version numbering may still be a bit WET. Working on reducing that in the future...

eddieantonio: Dumb question: why is there a separate file version and a lexical model version. Should they not be the same?

mcdurdin: Technically, the lexical model is developed independently to the package that contains it. In this repo, it makes sense for the .model.kps version to follow the lexical model version precisely... we have that functionality available with the `<FollowKeyboardVersion />` property in the `<Options>` key -- which is badly named but too late to fix.

Looking at the compiler, it understands that property but does not yet use it ... so that's something we need to implement.

Issue #19, opened 2019-03-30
Move lexical model compiler to keyman repo


mcdurdin:
So. A few things:

1. The merge of the model info should warn if there are mismatching version numbers, and ...
2. The version number shouldn't be in the .model_info, and ...
3. The canonical data is always in the source files, not the .model_info, because ...
4. Lexical models can be built, stored and distributed outside the `lexical-models` repository and CI chain. Those lexical models won't have a .model_info file, because that is used only for the CI deployment. It is important we don't end up building too many dependencies on the repository structure by accident because it is likely, if we are successful, that this repo will get very large. The intent is for models in this repo to be of high release-level quality and we generally want to discourage experimental models here, because models here can be automatically installed by Keyman. This leads me to conclude that ...
5. The compiler needs to be moved out of the lexical-models repo and into the keyman repo, and the .model_info management needs to be clearly delineated. Now, the compiler is part of the Keyman Developer toolchain. So I think what we need to do is update the build process to pull the latest stable (or `$tier`) version of the compiler from downloads.keyman.com (so that's something that needs to be added to the Keyman Developer CI). We *could* register an NPM package but I don't think we are quite ready to do that; happy to defer to wiser heads though. For consistency, then ...
6. This same process needs to be done with the keyboards repo -- so we don't include the kmcomp compiler in the repo, rather pull it at first build (or at appropriate times) from downloads.keyman.com.

_Originally posted by @mcdurdin in https://github.com/keymanapp/lexical-models/pull/14#issuecomment-477800093_

eddieantonio:
I'm starting to feel the pain of leaving this unaddressed. I find myself needing to make changes both to the LMLayer (keymanapp/keyman) and the compiler _at the same time_. I think these codebases belong in the same repo. Proposal:

 1. Move `tools/` to `keymanapp/keyman/common/lexical-model-compiler`.
 2. Have this repo have a dependency in `package.json` to install the lexical model compiler, as `npm install github:keymanapp/keyman#master`. [We don't need to publish an NPM package](https://docs.npmjs.com/cli/install)!
 3. In this repo, have the `build.sh` automatically checkout the latest version of the compiler. Call the compiler, like `kmlmc` (Keyman Lexical Model Compiler) from this repo.

That's it! Thoughts @mcdurdin?

mcdurdin:
> 1. Move `tools/` to `keymanapp/keyman/common/lexical-model-compiler`.

[...]

PR #27, opened 2019-05-03
Implement new-style word breaking

eddieantonio:
This generates model code compatible with keymanapp/keyman#1755. The main idea here is that the `model.ts` can specify a string for `wordBreaking`, e.g., 

```ts
(new LexicalModelCompiler).compile({
  format: 'trie-1.0',
  wordBreaking: 'ascii',
  sources: ['example.tsv']
});
```

And the compiler will generate the code to instantiate the model.

`example.en.wordlist` has been changed to use this feature and compile with a "built-in" word breaker.

**NOTE**: I strongly suggest tidying up `index.ts` once this PR is merged before adding any more features.

### DIFF
--- a/release/example/en.wordlist/source/example.en.wordlist.model.kps
+++ b/release/example/en.wordlist/source/example.en.wordlist.model.kps
@@ -11,7 +11,7 @@
     <Name URL="">Example (English) Template Wordlist Model</Name>
     <Copyright URL="">© 2019 SIL International</Copyright>
     <Author URL="mailto:marc@keyman.com">Marc Durdin</Author>
-    <Version>0.1.0</Version>
+    <Version>0.1.1</Version>
   </Info>
   <Files>
     <File>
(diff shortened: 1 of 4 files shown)


jahorton:
What sort of issues are you noticing re: "tidying up `index.ts`"?  It'd help in creating an issue and making sure we handle that correctly.

eddieantonio:
Clocking in at over 150 lines, [LexicalModelCompiler.compile()](https://github.com/keymanapp/lexical-models/blob/master/tools/index.ts#L19-L230) is definitely showing symptoms of [long method](https://refactoring.guru/smells/long-method).

I think #19 is a priority over internal refactoring though. I feel like there will be more shotgun surgery where there will have to changes that affect both repositories.

Also, I think the [`switch` statement](https://github.com/keymanapp/lexical-models/blob/master/tools/index.ts#L111-L135) may be more maintainable as a polymorphic call to a "model builder" or something like that, that knows how to produce code for a specific model.

We also are lacking test for all of this code! 😱 

My philosophy is to tidy up before piling on more and more code, so I just feel more comfortable this way.

mcdurdin:
I'm with Eddie on the quality before quantity approach, at least in theory ;-) It can be a difficult path to maintain at times when we have seriously constrained resources!

The compiler code I originally wrote as a bootstrap to get things moving so I am not in any way attached to it, its design or even the philosophy of how it works. Do with it what you will!

PR #34, opened 2019-05-23
[str.trie_sencoten] Fix the model ID

darcywong00:
I was having issues on the mobile app installing example.str.trie_sencoten.model.kmp.

The model filename is `example.str.trie_sencoten.model.js`
so the lexical model ID `example.str.trie` is missing the `_sencoten` portion.

Note, the model will still be unavailable from api.keyman.com because of issue keymanapp/keyman#1793


mcdurdin:
Note, this should have been falling over in compile phase due to this ... I'd like to check this out before merging this PR.

mcdurdin:
See https://github.com/keymanapp/keyman/pull/1799 for validation PR

PR #38, opened 2019-05-28
[example.str.trie_sencoten] Implement key function for SENĆOŦEN.

eddieantonio:
This adds a key function that nullfies the difference between uppercase and
lowercase letters, and whether accents matter when typing SENĆOŦEN.

### DIFF
--- a/release/example/str.trie_sencoten/source/model.ts
+++ b/release/example/str.trie_sencoten/source/model.ts
@@ -1,12 +1,33 @@
-import LexicalModelCompiler from "@keymanapp/developer-lexical-model-compiler"; 
+import LexicalModelCompiler from "@keymanapp/developer-lexical-model-compiler";
 
 (new LexicalModelCompiler).compile({
   format: 'trie-2.0',
-  wordBreaking: {
-    allowedCharacters: { initials: 'abcdefghijklmnopqrstuvwxyz', medials: 'abcdefghijklmnopqrstuvwxyz', finals: 'abcdefghijklmnopqrstuvwxyz' },
-    defaultBreakCharacter: ' '
-  },
-  //... metadata ...
+  wordBreaking: 'default',
   sources: ['saanich.tsv'],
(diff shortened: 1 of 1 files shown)


release/example/str.trie_sencoten/source/model.ts
```diff
@@ -1,12 +1,33 @@
...
-  wordBreaking: {
-    allowedCharacters: { initials: 'abcdefghijklmnopqrstuvwxyz', medials: 'abcdefghijklmnopqrstuvwxyz', finals: 'abcdefghijklmnopqrstuvwxyz' },
-    defaultBreakCharacter: ' '
-  },
-  //... metadata ...
+  wordBreaking: 'default',
   sources: ['saanich.tsv'],
-  // TODO: could include source code for key function here:
+  searchTermToKey: function (wordform: string): string {
+    const CEDILLA = '¸';
```
darcywong00: Would it be good to also include this?
```
const COMBINING_CEDILLA = '\u0327';
```

(Interesting on the GitHub review how the cedilla looks like a period. I pasted it elsewhere and it looked right)

eddieantonio:
It would not: the cedilla is not used as a diacritical mark, but as a full
letter in itself, representing the glotal stop. So the combining cedilla is
not used for anything, to my knowledge.

See the https://help.keyman.com/keyboard/fv_sencoten/9.1/fv_sencoten.php

Sometimes people use a comma instead of a cedilla! 🤪

release/example/str.trie_sencoten/source/model.ts
eddieantonio: It would not: the cedilla is not used as a diacritical mark, but as a full
letter in itself, representing the glottal stop. So the combining cedilla is
not used for anything, to my knowledge.

See the https://help.keyman.com/keyboard/fv_sencoten/9.1/fv_sencoten.php

Sometimes people use a comma instead of a cedilla! 🤪

PR #41, opened 2019-06-05
nrc.str.sencoten 1.0.1

eddieantonio:
I cleaned up Tim's word list a little bit, using the attached Python 3.6 program. Notably, this removes some “duplicate” suggestions like `TŦE` and `TŦE…`.


eddieantonio:
@darcywong00 Interesting! Maybe file a bug for me and @jahorton to look into in keymanapp/keyman?

PR #42, opened 2019-06-06
[example.en.mtnt] Add English model based on MTNT data.

eddieantonio:
I have created a full language model for the English language based on Reddit discussion posts. It comes from a text corpus collected by academics to stress-test machine translation models called [MTNT](http://www.cs.cmu.edu/~pmichel1/mtnt/).

Pros:
 - This describes [English as she is spoke](https://en.wikipedia.org/wiki/English_As_She_Is_Spoke) in contemporary internet discussions.

Cons:
 - The internet is a terrible place.

We can use this word model to practically test the predictive text feature, as well as discuss the ethics of data curation!


mcdurdin:
I think we will definitely need to cull pejorative words, racial or sexual slurs, etc. from the list before we can publish. I haven't had a good look at the list as yet, and I am aware that it can be tricky to determine where to draw the line but I don't think it would be wise to be offering up any of those words as suggestions.

Better to offend a few people by excluding certain words than to offend many by including them ;-)

Anyone want to take on that "fun" job? No doubt there is a list online somewhere we could use as a filter as a starting point.

jahorton:
>     * The internet is a terrible place.

😅

eddieantonio:
Okay! I removed everything present in MTNT's [`profanities.en`][prof] file (warning: creative profanities abound). I also did a brief skim of the data, trying to remove offensive words. I think the top 4000 are more-or-less okay. There may be still some objectionable stuff, but it's down to the point where I have to google for things I don't understand just to make sure if it's offensive or not :S 

So... yeah. I think what's left is infrequent enough that it will be difficult to get accidentally predicted anyway.

This data set is great if you're writing internet forum posts about Fortnite, or Game of Thrones 😂 

I need some eyebleech now. Here's a CC-BY picture of a kitten and a dog snuggling:

kitten and dog snuggling
> By [Bryan Alexander](https://www.flickr.com/photos/bryanalexander/4357340143)</small>


[prof]: https://github.com/pmichel31415/mtnt/blob/master/resources/profanities.en

eddieantonio:
Also, why is the build failing? It's claiming that `keyman/developer/package.json` doesn't exist...

```
npm WARN saveError ENOENT: no such file or directory, open 'C:\BuildAgent\work\2106c88466c099f3\keyman\developer\js\package.json
```

When it clearly does: https://github.com/keymanapp/keyman/blob/master/developer/js/package.json

rfaramir:
I noticed a bunch of numbers in it, which are pretty much useless words. Got rid of 662 of them in one blow by sorting by first field, deleting the first 662 lines, and re-sorting by second field.
I felt so clever until I noticed that words of the same count are now differently ordered, so will appear as unnecessary diffs. :(
Should I check it in anyway?

eddieantonio:
> I noticed a bunch of numbers in it, which are pretty much useless words. Got rid of 662 of them in one blow by sorting by first field, deleting the first 662 lines, and re-sorting by second field.
> I felt so clever until I noticed that words of the same count are now differently ordered, so will appear as unnecessary diffs. :(
> Should I check it in anyway?

That's fine. We don't really need a stable sort. I was thinking of doing a squash and merge rather than a regular merge for this one.

darcywong00:
imho, here's some more line numbers to redact from mtnt.tsv (I believe other variations were already removed)

7945
10508
11870
16002
21281
24291



Separate question, are we keeping emojis in the list?
(15057, etc)

eddieantonio:
> Separate question, are we keeping emojis in the list?
> (15057, etc)

I think the emoji frequencies are _interesting_, but not useful in this particular language model. I have ideas that I have to run across with @jahorton to make emoji prediction, which, as silly as it sounds, I think is a pretty big deal.

mcdurdin:
> Also, why is the build failing? It's claiming that `keyman/developer/package.json` doesn't exist...
> 
> ```
> npm WARN save
> Error ENOENT: no such file or directory, open 'C:\BuildAgent\work\2106c88466c099f3\keyman\developer\js\package.json
> ```
> 
> When it clearly does: https://github.com/keymanapp/keyman/blob/master/developer/js/package.json

Looking into this. I noticed it last night as well.

mcdurdin:
I can't see any reason why one build failed and the next one succeeded. This bothers me. I wondered if it was due to file locks from another process, but couldn't see any evidence of that. I will continue to monitor.

mcdurdin:
> I think the emoji frequencies are _interesting_, but not useful in this particular language model. I have ideas that I have to run across with @jahorton to make emoji prediction, which, as silly as it sounds, I think is a pretty big deal.

I agree that we should treat emoji as a separate thing. There are other ways of looking them up, etc, which we will probably need to spend some time on at some point. So would support removing them from this list.

mcdurdin:
> I need some eyebleech now. Here's a CC-BY picture of a kitten and a dog snuggling:

And thank you for taking on that task! It'll be an easier job in languages we don't understand ;-)

eddieantonio:
Hey, @rfaramir! I removed all digits in a stable way. @darcywong00, I also removed emoji — that's a task for another time! This one is ready for review!

rfaramir:
Thanks. I'll remove my attempt at cleaning it of numbers (it stayed local,
fortunately).
I look forward to more models!

PR #44, opened 2019-06-19
[example.str.trie_sencoten] remove model.

eddieantonio:
This model has been completly obsoleted by nrc.str.sencoten.

Since we're still in alpha, I decided to skip the deprectation cycle altogether to reduce confusion.


eddieantonio:
@mcdurdin it won't let me merge without the status check passing 😞. By the way, the status check failure makes no sense — it says `ENOENT: no such file or directory, open 'C:\BuildAgent\work\2106c88466c099f3\keyman\developer\js\package.json`?!?

mcdurdin:
Investigating why that file is missing sometimes ... it's happened before but we haven't pinpointed the issue yet. Re-ran the status check (on the same build agent) and that issue did not arise.

mcdurdin:
Going ahead and merging.

PR #47, opened 2019-07-02
[nrc.str.sencoten] better wordlist generation

eddieantonio:
I'm updating how the word list is generated from the raw sources given to me by Dr. Montler. There are definitely some big omissions and oddities in the original word list, so I expect to run `fix-wordlist` a few more times in the future! >.<


eddieantonio:
Ready for review! After direct correspondence with Dr. Montler, I figured out what was wrong. He further refined the list, and I added a few new words.

release/nrc/str.sencoten/source/Makefile
```diff
@@ -0,0 +1,10 @@
+# How to build the wordlist!
+#
+# Sources:
+#  - SaanichWordFreq.txt: Provided by Dr. Timothy Montler.
+#    NOT included in this repository!
+#    SHA384 sum: 5e31a32eaab3aa23e029f72fb3097f9fb97982fe9c927dd99093d3f1071c1b81d43926e4c9abc79a1afe75c63cfb599c
+SOURCES = SaanichWordFreq.txt
```
mcdurdin: If the source material is not available, then the Makefile doesn't really belong in the repo either -- it adds no value to other maintainers and it'd be better to keep that with the source data. And I guess that is also true of the fix-wordlist.py file?

The problem is that including these files puts the burden of maintaining them onto the repository maintainers -- but they have no ability to test the veracity of them or to do anything useful with them. If someone else (who presumably has access to the source data) submits a PR that updates them, the maintainers cannot actually test anything about the files so it becomes difficult to know whether to accept or reject the change.

If you do decide you want to keep the files in the repo, can you put them instead into `extra/` instead of `source/`, so it can be clear that they are not source files that anyone else can do anything useful with.

release/nrc/str.sencoten/source/saanich.tsv
```diff
@@ -1,8 +1,7 @@
 TŦE	13644
```
mcdurdin: I'm a little puzzled by this -- the only change I can see is a fairly minor change in frequency and the removal of one common word `ȻNES`?

release/nrc/str.sencoten/source/Makefile
eddieantonio: Okay! I'll put it in `extra/` It's useful for me, as a maintainer of the model :)

release/nrc/str.sencoten/source/saanich.tsv
eddieantonio: I've added a few words based on my interaction with Dr. Montler. I initially rejected all words with an "F", because "F" is not a native phoneme in SENĆOŦEN; however, I learned that it _is_ valid in loan words. He also cleaned up the tokenization of the original list, giving slightly better results.

"ȻNES"  is still there; it has just dropped significantly lower in the list. As written in the comments of `fix-wordlist.py`, "ȻNES" is not an orthographically correct word, but rather a common token typed by older L2 learners. Kind of like "would of" "could of" in English, in that it's incorrect, but it's widely used enough that it'd be annoying if it _wasn't_ there. So I let it be in the list with a relative frequency lower than "Ȼ" and "NE" which _are_ orthographically correct.

extras/nrc/str.sencoten/fix-wordlist.py
```diff
@@ -26,12 +26,16 @@ def uplus(c):
...
-        'I Í J K Ꝁ K̵ ₭ Ḵ Ḱ L Ƚ M '
-        'N Ṉ O P Q S Ś T Ⱦ T̸ Ṯ '
-        'Ŧ U W W̱ X X̲ Y Z ¸'
+    # Characters in the SENĆOŦEN alphabet:
+    'A Á Ⱥ B C Ć Ȼ D E H '
+    'I Í J K Ꝁ K̵ ₭ Ḵ Ḱ L Ƚ M '
+    'N Ṉ O P Q S Ś s T Ⱦ T̸ Ṯ '
+    'Ŧ U W W̱ X X̲ Y Z ¸'
+    # Characters used in loan words from English
+    'F'
```
eddieantonio: This addition will add words to the wordlist.

extras/nrc/str.sencoten/fix-wordlist.py
```diff
@@ -91,6 +105,22 @@ def uplus(c):
 
         wordlist[word] += count
 
-# Print as CRLF TSV:
-for word, count in wordlist.most_common():
-    print(word, count, sep='\t', end='\r\n')
+    # XXX: Artificially add ȻNEs
+    # It's not technically a single word according to the standard orthography
+    # e.g., *"ȻNEs YÁ¸" should be written "Ȼ NE SYA¸" ("that I go").
```
eddieantonio: This still adds "ȻNEs", but at a lower frequency than orthographically correct versions.

mcdurdin:
I'm sorry, miscommunication here: I meant put the extra files into `release/nrc/str.sencoten/extras/` :)

eddieantonio:
@mcdurdin Alright. Extras are in `elease/nrc/str.sencoten/extras` now :)

eddieantonio:
@mcdurdin I'd merge this, but the build is not working :( Could you restart the build or force it?

mcdurdin:
@eddieantonio Waiting on a number of other builds -- lots of keyboard work being done at present. I've bumped this up the queue for you.

mcdurdin:
Well, that went badly. Will need to investigate but won't be able to until tomorrow :(

mcdurdin:
That was a deeper rabbit hole than I expected. The developer lexical model compiler now fails to build index.ts because it isn't in the list of includes in tsconfig.json. Visible here because we are doing a clean build. This came out of changes in keymanapp/keyman#1861. Have pushed a new PR keymanapp/keyman#1871 to fix that, then should be able to re-run this build once that lands.

mcdurdin:
Okay, that PR fixed it, this is now ready to merge.

eddieantonio:
Yay! Thanks, @mcdurdin!

PR #48, opened 2019-07-03
[multiple] Model reorganization

### LINKED ISSUE #45: Rename or move example template models
I think as we head to beta we should be changing our example template models to use a custom BCP47 tag such as `qaa` so that they don't appear as options to end users, esp as we have an English model now. We definitely need to do this for release. That means a bunch of renaming though!

I would also like to see the English model move out of 'example' and into something real.

An alternative is to move the example models out of release/ into a sample/ folder. Perhaps that's better?

If we do it in alpha, we avoid the need to have backward compat and can just do cleanup on the downloads server. Once we go to beta, we have to deprecate and have a deprecation pathway.

eddieantonio:
This moves example models into `sample`, and moves `example.en.mtnt` to `nrc.en.mtnt`.

Closes #45.


eddieantonio:
@mcdurdin is TeamCity still be reconfigured?

mcdurdin:
> @mcdurdin is TeamCity still be reconfigured?

I have reset the build agent in question which still had a corrupted checkout of that folder. I've run out of time to figure out the root cause, but it may have been the misconfiguration I mentioned earlier which for some reason TC didn't clean up. Build should go through now, barring issues in this PR ;-)

mcdurdin:
I've also now set the required checks to reference the new tag `Test: Pull Requests (Models)`.

mcdurdin:
Okay, I have cleaned up the downloads server as well; only 2 NRC models listed there now, matching what we have reported from api.keyman.com.

PR #51, opened 2019-07-22
Add documentation: How to build a lexical model

eddieantonio:
Currently a work-in-progress...


eddieantonio:
So I still need to fill out some sections before this is finished, but I'd appreciate some preliminary comments from e.g., @jahorton, @mcdurdin, and @darcywong00. Writing this document has taught me that creating a lexical-model is still pretty difficult 😭

mcdurdin:
So ... not to make your life difficult or anything, but the documentation needs to make its way onto help.keyman.com which is where all the other documentation is. And needs to be in html (with php wrapper) format because I hate all life forms.

mcdurdin:
That said, fantastic work so far. I will review properly as soon as I can!

eddieantonio:
> So ... not to make your life difficult or anything, but the documentation needs to make its way onto help.keyman.com which is where all the other documentation is. And needs to be in html (with php wrapper) format because I hate all life forms.

Haha! I was asking people where it should go, and people said here. Sure, I'll migrate it over to help.keyman.com, but maybe I work on it a bit more here :)

docs/how-to-build-a-lexical-model.md
```diff
@@ -0,0 +1,365 @@
...
+Requirements
+------------
+
+You will need to have:
+
+ - [Node.js][] >= 10.0
+ - (Windows only) [Git Bash](https://gitforwindows.org/)
+
+
+[Node.js]: https://nodejs.org/en/
```
mcdurdin: Might as well put that link inline if you doing that for Git Bash anyway (but formatting will change when it moves...)

docs/how-to-build-a-lexical-model.md
```diff
@@ -0,0 +1,365 @@
...
+The first column **MUST** be the words, and the second column **MUST**
+be the counts. Additional columns are ignored. The spreadsheet can also
+be a single column of all of the words in the language.
+
+This is what my word list looks like in Google Sheets:
+
+![The word list, as it appears in Google Sheets](./sencoten-sheets-full.png).
+
+[sencoten-sheet]: https://docs.google.com/spreadsheets/d/10zhIc439BCSSooL_-HeJ6TUHd-ovkiXYcIGe-pHDTSg/edit?usp=sharing
+zsh:1: no matches found: https://docs.google.com/spreadsheets/d/10zhIc439BCSSooL_-HeJ6TUHd-ovkiXYcIGe-pHDTSg/edit?usp=sharing
```
mcdurdin: Is this supposed to be here? See also next line!

docs/how-to-build-a-lexical-model.md
```diff
@@ -0,0 +1,365 @@
...
+
+> _author_._bcp47_._uniq_
+
+
+### Author
+
+_author_ is an identifier for the maintainer of the model. This
+identifier is a string of one or more ASCII lowercase letters, digits, and the
+underscore (`_`), and it must begin with a lowercase ASCII letter. Some
+organizations have a author identifier already, like `sil` for SIL
```
mcdurdin:
```suggestion
organizations have an author identifier already, like `sil` for SIL
```

docs/how-to-build-a-lexical-model.md
```diff
@@ -0,0 +1,365 @@
...
+
+[ISO 639-3]: https://iso639-3.sil.org/code_tables/639/data
+[ISO 15924]: https://en.wikipedia.org/wiki/ISO_15924
+[BCP 47]: https://tools.ietf.org/html/bcp47
+
+### Uniq
+
+_uniq_ is an arbitrary tag that uniquely identifies your specific model.
+This identifier can be anything, as long as it hasn't already been used
+before for this specific language. As with `author`
```
mcdurdin: Uniq only needs to be unique to the author, for that language; it doesn't technically need to be unique to other people.

Also note that this file naming scheme is only mandated for models in the shared repo. For private distribution you should be able to call it `foo` (even if that isn't recommended!)

docs/how-to-build-a-lexical-model.md
```diff
@@ -0,0 +1,365 @@
...
+            └── wordlist.tsv
+```
+
+We need to create the remaining files:
+ - The "model info" file: `example.str.tutorial.model_info`
+ - The package file: `example.str.tutorial.model.kps`
+ - The model configuration file: `model.ts`
+
+
+Step 4.1: Creating the model info file
```
mcdurdin: model_info should only be required when building within the repo and distributing via keyman cloud (not used for private, ad-hoc distribution; worth noting this). We'll be adding outside-repo builds before release, I hope!

docs/how-to-build-a-lexical-model.md
```diff
@@ -0,0 +1,365 @@
...
+}
+```
+
+I'll save this file to `example/str.tutorial/example.str.tutorial.model_info`.
+
+
+Step 4.2: Creating the KPS file
+-------------------------------
+
+**TODO:**
```
mcdurdin: We'll generally encourage model developers to use Keyman Developer to create this file. It saves on silly errors. Clean support for models in Developer however is of course TBD.

docs/how-to-build-a-lexical-model.md
```diff
@@ -0,0 +1,365 @@
...
+
+Step 4.3: Creating the model file
+---------------------------------
+
+Finally, the `model.ts` is a TypeScript source code file that is
+executed to build the model. For our purposes, we can simply copy-paste
+the following example:
+
+```typescript
+import LexicalModelCompiler from "@keymanapp/developer-lexical-model-compiler";
```
mcdurdin: Also note: when we integrate as part of Keyman Developer, we'll be making this compiler available as part of the install process.

docs/how-to-build-a-lexical-model.md
```diff
@@ -0,0 +1,365 @@
...
+ - **comment** lines MUST start with the '#' character on the very first column
+ - **entries** are one or two columns, separated by the (horizontal)
+   **tab character** (U+0009)
+ - column 1 (REQUIRED): the **word form**: can have any character
+   except tab, CR, or LF. Surrounding whitespace characters are trimmed.
+ - column 2 (optional): the **count**: a non-negative integer specifying how many
+   times this entry has appeared in the corpus. Blank means
+   'indeterminate', and is treated as if the word exists in the corpus,
+   but will be predicted at the lowest possible priority.
+ - the following columns are ignored.
```
mcdurdin:
```suggestion
 - any other columns are ignored.
```

docs/how-to-build-a-lexical-model.md
```diff
@@ -0,0 +1,365 @@
...
+The **lexical model compiler** expects wordlists to abide by the
+following **tab-separated values** (TSV) format:
+
+ - the file is must be **UTF-8** encoded text file
+ - newlines are either **LF** or **CRLF**
+ - the file either consists of a **comment** or an **entry**
+ - **comment** lines MUST start with the '#' character on the very first column
+ - **entries** are one or two columns, separated by the (horizontal)
+   **tab character** (U+0009)
+ - column 1 (REQUIRED): the **word form**: can have any character
```
mcdurdin: Suggest noting that quote characters are not used to surround the text?

docs/how-to-build-a-lexical-model.md
```diff
@@ -0,0 +1,365 @@
...
+   'indeterminate', and is treated as if the word exists in the corpus,
+   but will be predicted at the lowest possible priority.
+ - the following columns are ignored.
+
+> Source:
+> [build-trie.js@e58173f](https://github.com/keymanapp/keyman/blob/307436e7c24caa5c720b272640184362c4dc3223/developer/js/lexical-model-compiler/build-trie.ts#L21-L66)
+
+Exporting a spreadsheet from Google Sheets as a TSV will produce
+properly formatted output. I have not found a reliable output option in
+Microsoft Excel for Mac that will produce output in the expected format.
```
mcdurdin: Excel for Windows can load the .tsv file okay., but Save has trouble with encoding. However, Copy/Paste into Notepad++ works (older versions of Notepad would work except they prepend a UTF-8 BOM 😭; new versions don't do this).

docs/how-to-build-a-lexical-model.md
```diff
@@ -0,0 +1,365 @@
...
+----------------------------------------------------------
+
+<a id="appendix:tsv"></a>
+
+[appendix]: ./#appendix:tsv
+
+The **lexical model compiler** expects wordlists to abide by the
+following **tab-separated values** (TSV) format:
+
+ - the file is must be **UTF-8** encoded text file
```
mcdurdin:
```suggestion
 - the file must be **UTF-8** encoded text file without Byte Order Mark (BOM).
```

Note: Notepad until recently always prepended a BOM. This is now changing: https://www.bleepingcomputer.com/news/microsoft/windows-10-notepad-is-getting-better-utf-8-encoding-support/

mcdurdin:
@DavidLRowe would you be willing to have a look at this as well? Given the work you are doing on the keyboards documentation, I thought this might have some overlap.

It will be migrating to help.keyman.com and format changing from Markdown to HTML, so don't worry about format-related stuff, just look at content.

mcdurdin:
Clarification point: I think we should be careful to separate model development from model distribution via Keyman Cloud (i.e. in the lexical-models repo). We've done some of the work now in separating those two by moving the lm compiler into the keyman repo, but we probably need to do more ;-)

I think I will work on the model integration in Keyman Developer as a matter of priority. As soon as we have the other code in beta!

docs/how-to-build-a-lexical-model.md
DavidLRowe: I'd find a more concrete example much more helpful. For example, Gene Rodgers of the National Research Council Canada is developing a lexical model for Straits Salish.
- author is `nrc` (for the National Research Council Canada)
- the BCP 47 code for the language is `str`
- the unique code is `grodgers` (or some other convention established by the NRC for uniquely identifying authors within the NRC)
- result is nrc.str.grdogers

Joe Bloggs is developing a lexical model for Hausa written with Arabic script.
- author is `bloggs`
- the BCP 47 code for the language is `ha_arab`
- the unique code is `joe` (or some other string that uniquely identifies Joe from other authors named `Bloggs`)
- result is bloggs.ha_arab.joe

docs/how-to-build-a-lexical-model.md
```diff
@@ -0,0 +1,365 @@
...
+            ├── wordlist.tsv
+            └── model.ts
+```
+
+Notice how under `release`, there is a folder that is named after the
+author. Within that folder is a folder for each models created by that
+particular author. Within each model folder, this is a folder called
+`source` that contains **wordlist.tsv**. We will discuss what the other
+files are called a bit later in this document.
+
```
DavidLRowe: change "each models" to "each model"

"this is a folder..." Did you mean "there is a folder..."?

docs/how-to-build-a-lexical-model.md
mcdurdin: That's not quite right :)

The author is already unique -- if two developers named Bloggs turned up wanting to use the `bloggs` subtag, we'd ask the second one to pick a different author subtag, (so the second developer may get `alfredbloggs` for the author subtag. The `uniq` subtag is really about disambiguating multiple models for the same language from the same author, so for example:

Joe Bloggs has a Straits Salish model based on the Montler wordlist, and another one based on the Salish School Board wordlist (this is semi-fictional but ...). He wants both of them available because of different preferences in the community. So he picks a uniq subtag for each accordingly, for example:

* `bloggs.str.montler`
* `bloggs.str.schoolboard`

In practice, I hope there won't be many models for the same language. Note that dialect differences should really be picked up in the BCP 47 subtag (e.g. en_us, en_ca, en_au)!

DavidLRowe: I appreciate the clarification. I'm still trying to wrap my head around this, and a concrete example helps -- especially when it shows that my understanding was wrong!!

mcdurdin: No problemo. The intent is to try to avoid running into namespacing issues which many projects (including keyman keyboards) have run into in the past. This may be a little over the top in some respects but isn't too hard, hopefully, for people to get their head around. The `uniq` subtag could also just be described as a name for the model.

docs/how-to-build-a-lexical-model.md
jahorton: I mean, is it hard to strip out the BOM?  Seems like a big ask for those not so familiar with byte encoding particularities.

docs/how-to-build-a-lexical-model.md
eddieantonio: haha, I think I keysmashed in Vim here...

docs/how-to-build-a-lexical-model.md
eddieantonio: Could you please elaborate on this point?

docs/how-to-build-a-lexical-model.md
eddieantonio: Correct!

docs/how-to-build-a-lexical-model.md
eddieantonio: Good point! It's not really that difficult, and I don't _think_ there's any problem here. I don't think... please debate it over here: keymanapp/keyman#1916!

docs/how-to-build-a-lexical-model.md
mcdurdin: Once we have Keyman Developer capable of doing lexical model builds, there won't be a need to hand-write all parts of these files. It will include a template for a model file to start people off and will make sure the compiler is readily available without needing to clone the repo.

I guess the comment was more a reflection on the current reality that that import statement requires a clone of the keymanapp/keyman repo which is somewhat less than ideal!

mcdurdin:
I think those changes look good. So next step is to abandon all this work here and start again in the help repo ;-)

eddieantonio:
> I think those changes look good. So next step is to abandon all this work here and start again in the help repo ;-)

Okay, I'll keep this one open until something is merged into help.keyman.com, as already people are referring to this incomplete documentation for the beta!

PR #58, opened 2019-10-07
Use lexical-model-compiler npm packages

darcywong00:
Now that Keyman 12.0 is released, we can update this repo to use the npm dependencies.
@keymanapp/lexical-model-compiler
@keymanapp/lexical-model-types

I've modified `-no-npm` flag to skip all npm steps. 

For my personal testing of `-no-npm`, I uninstalled Developer and 
```
npm uninstall -g @keymanapp/lexical-model-compiler
```

### DIFF
--- a/resources/compile.sh
+++ b/resources/compile.sh
@@ -145,16 +145,16 @@ function build_release_model {
   pushd source
 
   # Compile model
-  kmlmc -o "../build/$modelOutputFilename" "./$modelInputFilename" || die "Unable to build .model.js file"
+  npx kmlmc -o "../build/$modelOutputFilename" "./$modelInputFilename" || die "Unable to build .model.js file"
 
   # Compile package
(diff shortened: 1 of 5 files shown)


resources/compile.sh
```diff
@@ -145,16 +145,16 @@ function build_release_model {
   pushd source
 
   # Compile model
-  kmlmc -o "../build/$modelOutputFilename" "./$modelInputFilename" || die "Unable to build .model.js file"
+  npx kmlmc -o "../build/$modelOutputFilename" "./$modelInputFilename" || die "Unable to build .model.js file"
```
eddieantonio: FYI, another way to do this is:

```bash
$(npm bin)/kmlmc -o "../build/$modelOutputFilename" "./$modelInputFilename" || die "Unable to build .model.js file"
```

PR #60, opened 2019-11-21
Add sample tamil wordlist model

darcywong00:
This is an example/sample Tamil wordlist model to use with mobile Sample apps.

It builds fine with Developer. But `./build.sh` fails with a cryptic error
```
Building model sample/example/example.ta.wordlist/
Cannot read property 'match' of undefined
Unable to merge .model_info file
Aborting with error 1
```


eddieantonio:
I think it's an error in `kmlmi`: https://github.com/keymanapp/lexical-models/blob/543464cc0d421f432122bb766d771a4e26d856bc/resources/compile.sh#L157-L164

It seems hardcoded to expect an author `mailto:` URL in the KPS file, and you have nothing defined in your .kps file: https://github.com/keymanapp/keyman/blob/3ce626d583359a57d33ef3f75adc9db040e4ac96/developer/js/source/model-info-compiler/model-info-compiler.ts#L89-L90

This seems like a bug. As a workaround, add any `mailto:` url in the Author email field in the `.kps`.

sample/example/example.ta.wordlist/source/example.ta.wordlist.model.kps
```diff
@@ -0,0 +1,40 @@
...
+  </Options>
+  <StartMenu>
+    <Folder></Folder>
+    <Items/>
+  </StartMenu>
+  <Info>
+    <Name URL="">Example (Tamil) Template Wordlist Model</Name>
+    <Version>1.0</Version>
+    <Copyright URL="">© 2019 SIL International</Copyright>
+    <Author URL="">Darcy Wong</Author>
```
eddieantonio: Please add an email (even a dummy one like `mailto:nobody@example.com`) here.

darcywong00: ooh, thanks for pointing that out! It builds now.

PR #64, opened 2020-01-15
Initial version of sil.bcc-arab.upp_ptwl1

rmlockwood:

PR #65, opened 2020-01-21
Update lexical-model-compiler

darcywong00:
Update the version of lexical-model-compiler to `12.0.66` so #64 will build.


mcdurdin:
Note you can also do `npm update` to get past this

PR #71, opened 2020-02-27
Feat/unilex

### LINKED ISSUE #70: Build default lexical models from Unicode unilex data
https://github.com/unicode-org/unilex/tree/master/data/frequency

I reckon we could get a long way with default models. Not perfect for all languages but maybe a decent base for others to work on.

Also they appear to be TSV files -- just need to strip off a line or two at the start!

mcdurdin:
These models are generated automatically and no guarantees as to quality.
This is an experiment -- let's discuss!

Source data is from https://github.com/unicode-org/unilex

Fixes #70.


mcdurdin:
Stale, could use Unilex as source data but no evidence that it is where we want to go at this time.

PR #74, opened 2020-02-27
[lex-jv] Initial version 1.0

bennylin:
New Javanese wordlist


keyman-server:

darcywong00:
Just checking about wordlist.tsv:

* There might be some repeats (for example: `aku`  lines 12, 530, 547, 824)

bennylin:
I tested it and it worked fine, so I didn't delete the duplicates. Would that pose a problem?

Also, there are some phrases in there as well. Do I need to clean them up into 'words'-only list?

jahorton:
I'd have to investigate to be sure on these points, since the behavior isn't explicitly defined, but I'd expect the following behaviors:

- For phrases:  they may show up as suggestions if they can be matched on the first word, but the second you start typing the second word of the phrase, it can no longer show up as a suggestion.  (We don't have anything to properly handle phrases yet.)
- For repeating the same word in the wordlist:  as far as I can tell, this will probably cause issues with the frequency counts, and I'm not sure what determines which frequency wins.  We use that to determine what suggestions are more likely than others, so it may affect suggestion quality.

darcywong00:
> For repeating the same word in the wordlist: as far as I can tell, this will probably cause issues with the frequency counts, and I'm not sure what determines which frequency wins. We use that to determine what suggestions are more likely than others, so it may affect suggestion quality.

Yeah we still recommend deleting the duplicates in the wordlist.

PR #80, opened 2020-03-13
[wakhi] submission on behalf of author

DavidLRowe:


darcywong00:
We recommend the wordlists get cleaned up to remove "duplicates" and use NFC instead of NFD.

Here's an example of a "duplicate"
```
ав̌лӣ  4
ав̌лӣ  1
```
The wordcount 4 ends with U+0438 U+0304 (Cyrillic small letter i + combining macron)
The wordcount 1 ends with U+04E3 (Cyrillic small letter i with macron)

PR #93, opened 2020-05-02
[example.en.custom_word_breaker]: new model

eddieantonio:
Create a new, fully-working sample model that provides its own custom word breaker function.

Closes keymanapp/keyman#2048


jahorton:
This would give us a second `en` lexical model.  Did we ever resolve how the API chooses a preferred model?  We should make sure that's properly handled before this gets merged.

eddieantonio:
> This would give us a second `en` lexical model. Did we ever resolve how the API chooses a preferred model? We should make sure that's properly handled before this gets merged.

But will this be a published model? It's in the `samples` directory, so I assume it's not built/published by default...?

PR #96, opened 2020-07-03
cabecar predictive text model

victoriaq22:
lexical model for cabecar keyboard, wordlist updated to more accurately reflect word counts


keyman-server:

PR #104, opened 2020-11-04
[newa] Update node version and rebuild Newa model

### LINKED ISSUE #97: [newa] Newa autocomplete not working
Not getting proper suggestions after newa wordlist 2.0 is installed. I get only this suggestion (  𑐺𑐕𑐣𑑂𑐟 ) when one letter is typed, (regardless of what is typed) and no suggestions when more than one is typed.
distributed

However if i build locally and then install, suggestions are shown as expected.
local-build

[this is the build that works](https://sapradhan.keybase.pub/newa/assets/sapradhan.new-newa.nepal_bhasa.model.kmp?dl=1)

darcywong00:
Fixes #97 

The Newa model uses SMP characters.

The current version of node in package.json `^11.15.0` generates an invalid Newa lexical-model (splits surrogate pairs).
Updating node to the latest LTS `14.15.0` fixes the compilation.

Also bumped the Newa model version to force a rebuild.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -85,9 +85,9 @@
       }
     },
     "node": {
-      "version": "11.15.0",
-      "resolved": "https://registry.npmjs.org/node/-/node-11.15.0.tgz",
-      "integrity": "sha512-Nbzq8qr133iwjGo0ZtzQR0mYeawW2eddYpW/k/+yjgbQW2/zG1a/5QiizVOrJ8yc4hbu3369zkj4dDIEd8f7dg==",
+      "version": "14.15.0",
+      "resolved": "https://registry.npmjs.org/node/-/node-14.15.0.tgz",
+      "integrity": "sha512-FrsP5wcA72CXNgQUk7zIdZm4vciBa/ahzaGC5iv3T0coNvz7hGsiI4pMdqqr0OXlVqyvSxDHzUUrhxlY3Hb2Kg==",
       "requires": {
         "node-bin-setup": "^1.0.0"
       }
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "@keymanapp/lexical-model-compiler": "^13.0.109",
     "@keymanapp/lexical-model-types": "^13.0.107",
     "@types/node": "^10.17.27",
-    "node": "^11.15.0",
+    "node": "^14.15.0",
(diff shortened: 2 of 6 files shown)


jahorton (approved):

mcdurdin:
It may be worth rebuilding a few of the other models and comparing the output to ensure the Node version upgrade doesn't break something else.

darcywong00:
> It may be worth rebuilding a few of the other models and comparing the output to ensure the Node version upgrade doesn't break something else.

I did a full rebuild on the repo and compared it to the master branch.
They only model I noticed a diff was benny_lin.id.kamus_indonesia.model.js where the "values" array of the same weight were listed in different order
e.g.
```
{"type":"internal","weight":1,"values":["﷐","t","a","g","-",null]
vs
{"type":"internal","weight":1,"values":["﷐","t","g","-","a",null]
```

The "broken" character is U+FDD0 : <noncharacter>. I don't know why that particular model is generating that, but that's a separate thing.

eddieantonio:
> > It may be worth rebuilding a few of the other models and comparing the output to ensure the Node version upgrade doesn't break something else.
> 
> I did a full rebuild on the repo and compared it to the master branch.
> They only model I noticed a diff was benny_lin.id.kamus_indonesia.model.js where the "values" array of the same weight were listed in different order
> e.g.
> 
> ```
> {"type":"internal","weight":1,"values":["�","t","a","g","-",null]
> vs
> {"type":"internal","weight":1,"values":["�","t","g","-","a",null]
> ```
> 
> The "broken" character is U+FDD0 : . I don't know why that particular model is generating that, but that's a separate thing.

The U+FDD0 is by design. See: https://github.com/keymanapp/keyman/blob/573c63a931a7bd9fca8618e15044190947380aaa/developer/js/source/lexical-model-compiler/build-trie.ts#L274-L284

DavidLRowe:
release\sapradhan\sapradhan.new-newa.nepal_bhasa\README.md  needs version number changed to 2.2

darcywong00:
> The U+FDD0 is by design

Should the lm compiler output `\ufdd0` instead?

mcdurdin:
> Should the lm compiler output `\ufdd0` instead?

No, it's fine to output the character as is.

PR #105, opened 2020-11-07
SGW ⇒ SGW-ETHI

dyacob:
The change of language identifier is made to synchronize with the identifier used in the GFF Gurage keyboard, and is discussed in the Keyman App Issue 3806 (https://github.com/keymanapp/keyman/issues/3806).

While this change is necessary, one issue does arise whereby this lexical model uses the provisional Gurage orthography applied by the GFF Gurage (gff_gurage) keyboard, but  **_not_** by the GFF Gurage Legacy (gff_gurage_legacy) keyboard.  However, both keyboards use the sgw-Ethi identifier.  

Is there any change that could be made to the respective identifiers such that this lexical model is **_only_** used by the gff_gurage keyboard?

### DIFF
--- a/release/gff/gff.sgw.gurage/HISTORY.md
+++ b/release/gff/gff.sgw.gurage/HISTORY.md
@@ -1,6 +1,10 @@
 Gurage Lexical Model Change History
 ====================
 
+1.01 (2020-11-07)
+----------------
+* Change of language identifier to sgw-Ethi to synch with keyboards.
+
 1.0 (2020-02-26)
 ----------------
 * Initial Release with UniLex, Wolf Leslau, and Sahle Jingo word lists.
--- a/release/gff/gff.sgw.gurage/README.md
+++ b/release/gff/gff.sgw.gurage/README.md
@@ -3,12 +3,12 @@ Gurage Lexical Model
 
 © 2020 Ge'ez Frontier Foundation
 
-Version 1.0
+Version 1.01
 
 Description
 -------------
 
-This is a Gurage (ጉራጌ , ISO-639-2 sgw) lexical model developed to support predictive text the
+This is a Gurage (ጉራጊና, ISO-639-2 sgw) lexical model developed to support predictive text the
 corresponding GFF Gurage Keyboard.  The lexical model comprises word frequency lists from
 the works of Wolf Leslau ("Ethiopians Speak vol. 2 Chaha"), Shahle Jingo (Gurage Fables)
 and the Sebat Bet Bible of the Ethiopian Bible Society. The respective texts have been normalized
--- a/release/gff/gff.sgw.gurage/gff.sgw.gurage.kpj
+++ b/release/gff/gff.sgw.gurage/gff.sgw.gurage.kpj
@@ -19,12 +19,12 @@
       <ID>id_60dbd3dc5598c8a7525f9f53eae22083</ID>
       <Filename>gff.sgw.gurage.model.kps</Filename>
       <Filepath>source\gff.sgw.gurage.model.kps</Filepath>
-      <FileVersion>1.0</FileVersion>
+      <FileVersion>1.01</FileVersion>
       <FileType>.kps</FileType>
       <Details>
         <Name>Gurage</Name>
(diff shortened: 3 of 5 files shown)


keyman-server:

LornaSIL:
I think Windows still requires you to use sgw-Ethi...but for the legacy one you could add an -x subtag. It needs to be eight or fewer characters. Using that would prevent the lexical model to be used for legacy.
I'd like to discuss it with you before you decide for sure because we are recording these in our language tags database:
sgw-Ethi-ET-x-bse (for the orthography used by Bible Society of Ethiopia)
sgw-Ethi-ET-x-sebatbet (for the preferred language name at the time)
sgw-Ethi-ET-x-1990 (for whatever year it was used)

The subtag would be sgw-Ethi-x-bse (etc.)

You CAN use digits (like the 1990 example), but I believe that is not preferred.

dyacob:
Hi @LornaSIL,

I was just trying to change the language tag (BCP 47 Tag) to use one of the -x- subtags but they were rejected, with the message "...must contain only Language-Script-Region".  

I was modifying the the &lt;Language ID="sgw-Ethi"&gt; markup in the .kps file.  Is there somewhere else to set the subtag?

My preference would be to use sgw-Ethi-ET-unicode-4.1  to be politically neutral.  Or something like this, -x-uni41 minimally if the extra "." and "-" would be rejected.

thanks!

DavidLRowe:
We might need @mcdurdin to advise us.

dyacob:
Do any keyboards presently use a -x- subtag that could serve as a working example?

DavidLRowe:
I did a quick scan of .keyboard_info and .kps files and found none that use -x-.

mcdurdin:
There are some platforms that have trouble with BCP 47 tags that contain subtags outside of the lang-script-region triplet, so we are not currently recommending using those. That's not ideal, and my preference would be for Keyman on those platforms to massage the BCP 47 tags into something that is acceptable on that platform, but we're not there yet.

Having a complex BCP 47 tag does also pose some complications currently for the matching between keyboards and lexical models. This is something that still needs work also.

dyacob:
Could I give the "gff_gurage_legacy" keyboard an artificial region named "Legacy"?  Thus the language identifier "sgw-Ethi-Legacy"?   Alternatively, I could use a real region for one keyboard (and lexical file) but not the other.

DavidLRowe:
There are private use region codes.

From the BCP 47 spec (https://tools.ietf.org/html/bcp47#section-2.2.4):
```
The region subtags 'AA', 'QM'-'QZ', 'XA'-'XZ', and 'ZZ' are reserved for private use in language tags. 
These subtags correspond to codes reserved by ISO 3166 for private use.  
These codes MAY be used for private use region subtags (instead of using a private use subtag sequence).
```
Further details are in https://tools.ietf.org/html/bcp47#section-4.6

So `sgw-Ethi-XE` perhaps? Not sure how that will play in the Keyman ecosystem, but at least it fits the BCP 47 spec.

dyacob:
Thanks @DavidLRowe .  I went with "XL" where "L" is mnemonic for "Legacy".  So what we arrive at is then:

### sgw-Ethi-XL:
-  gff_gurage_legacy (keyboard)

### sgw-Ethi:
 -  gff_gurage  (keyboard)
 -  gff.sgw.gurage (lexical model)

I'm assuming with this convention when the `gff_gurage_legacy` keyboard is installed, it will not attempt to then retrieve and install the `gff.sgw.gurage` lexical model, which leads to the double menu entry problem recorded in: https://github.com/keymanapp/keyman/issues/3806 . @darcywong00 can you confirm?

mcdurdin:
We've not got enough runway left for Keyman 14.0 for us to resolve the remaining issues with BCP 47. Each release brings us slightly closer to where I would like to be. For now, let's go with the private region code (although I don't know if that is going to cause grief on any platforms), and in the future we'll aim to be supporting the `-x-variant` pattern which is much cleaner.

DavidLRowe (approved):

PR #106, opened 2020-11-26
glk-arab is an RTL script

jahorton:


darcywong00:
You'll need to bump the version (.kps, HISTORY.md) to force the package to build.

DavidLRowe:
Is the change to .gitignore intended? or did that sneak in somehow?

jahorton:
> 
> 
> Is the change to .gitignore intended? or did that sneak in somehow?

Both, actually.  That targets a file that macOS makes that has a tendency to sneak in otherwise.

DavidLRowe (approved):

PR #121, opened 2021-02-26
fix: updates compiler, MTNT due to de-duplication bug

jahorton:
This PR is motivated by https://github.com/keymanapp/keyman/pull/4504.

In short, a compiler bug partially broke the MTNT model's weighting.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,20 +5,20 @@
   "requires": true,
   "dependencies": {
     "@keymanapp/lexical-model-compiler": {
-      "version": "14.0.226-beta",
-      "resolved": "https://registry.npmjs.org/@keymanapp/lexical-model-compiler/-/lexical-model-compiler-14.0.226-beta.tgz",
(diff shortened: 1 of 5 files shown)


DavidLRowe:
I don't fully understand what this PR is doing, but I'm thinking the version number needs to change in the .kps file. (And might not have needed to change in the .kpj file? or would have been update automatically?)

jahorton:
> I don't fully understand what this PR is doing, but I'm thinking the version number needs to change in the .kps file. (And might not have needed to change in the .kpj file? or would have been update automatically?)

I _thought_ there was another spot I needed to update, but I couldn't remember it.  Thanks!

mcdurdin:
> And might not have needed to change in the .kpj file

Yes, .kpj files are updated automatically; the metadata in there is only for presentation purposes within the IDE.

mcdurdin (approved):

jahorton:
Yeah, apologies.  My Mac is my main development machine, and I don't have WINE or anything set up on it.  As a result, minor edits like this, I tend to to outside of Developer... hence the mistake.  Thanks again for catching that.

PR #122, opened 2021-03-04
[brb] Lave (Brao) spoken in Cambodia

MakaraSok:
Brao wordlist


release/sil/sil.brb.brao/sil.brb.brao.model_info
```diff
@@ -0,0 +1,7 @@
+{
+    "license": "mit",
+    "languages": [
+        "brb"
```
darcywong00, line 4: Maybe also include `brb-khmr` (here and .model.kps) so this model can be used with sil_khmer
https://keyman-staging.com/keyboards/sil_khmer?bcp47=brb-khmr#expand-languages

MakaraSok:
Adding the `brb-khmr` to the package causes a known error to arise
```
sil.brb.brao.model.kps: Compiling package sil.brb.brao.model.kps...
sil.brb.brao.model.kps: Warning: Lexical model with ID 'sil.brb.brao' has a BCP 47 warning: 'brb-Khmr' is a valid tag but is not canonical: it should be 'brb'.
sil.brb.brao.model.kps: Failure: 'C:\keyman\lexical-models\release\sil\sil.brb.brao\source\sil.brb.brao.model.kps' was not compiled successfully.
```

darcywong00:
Ah, we might need to update the lexical model compiler with keymanapp/keyman#4563 so that "error" becomes an "info"

MakaraSok:
related: https://github.com/keymanapp/lexical-models/issues/126

MakaraSok:
> Adding the `brb-khmr` to the package causes a known error to arise
> 
> ```
> sil.brb.brao.model.kps: Compiling package sil.brb.brao.model.kps...
> sil.brb.brao.model.kps: Warning: Lexical model with ID 'sil.brb.brao' has a BCP 47 warning: 'brb-Khmr' is a valid tag but is not canonical: it should be 'brb'.
> sil.brb.brao.model.kps: Failure: 'C:\keyman\lexical-models\release\sil\sil.brb.brao\source\sil.brb.brao.model.kps' was not compiled successfully.
> ```

@jahorton this error still appears when compiling. When will this be fixed? Or has this already been fixed, but I have not got the latest update on my master?

MakaraSok:
Further testing needed: when it comes to a consonant + a subscript, the model of this build does work when typing the sequence at the beginning of the line, but after picking a word from the suggestion and try the sequence again, the suggestions given are not the expected ones.

Screenshot_1617089650

jahorton:
> Further testing needed: when it comes to a consonant + a subscript, the model of this build does work when typing the sequence at the beginning of the line, but after picking a word from the suggestion and try the sequence again, the suggestions given are not the expected ones.

So... is this after backspacing to remove characters from an accepted suggestion, or is this when typing a new word after having recently accepted a suggestion?  I'm not fully clear on which scenario you mean here.

MakaraSok:
This is when typing a new word after having recently accepted a suggestion.

darcywong00:
This repo will need the lexical model compiler update in keymanapp/keyman#4892 (14.0.272 when it goes out)

jahorton:
> Further testing needed: when it comes to a consonant + a subscript, the model of this build does work when typing the sequence at the beginning of the line, but after picking a word from the suggestion and try the sequence again, the suggestions given are not the expected ones.
> 
> Screenshot_1617089650

Spent some time chasing this down; when I get everything properly set up, the error does not occur for me.

That said, let me emphasize:  **_when I get everything properly set up_**.  That's the tricky part here.  The copy of `sil_brao` I have appears to be set for language code `brb`, as opposed to this model's `brb-khmr`.  That's a mismatch - so the older copy of this model I had laying around, which was installed for `brb`, remained in-place.  That copy does produce this error.  We should probably double-check the language codes for keyboard and lexical model and ensure they properly match.

darcywong00:
> We should probably double-check the language codes for keyboard and lexical model and ensure they properly match.

A search for https://keyman.com/keyboards?q=l:id:brb

yields sil_khmer (includes brb-khmr)

jahorton:
> > We should probably double-check the language codes for keyboard and lexical model and ensure they properly match.
> 
> A search for https://keyman.com/keyboards?q=l:id:brb
> 
> yields sil_khmer (includes brb-khmr)

Yes, but noticing the screenshot @MakaraSok shared:

> Screenshot_1617089650

That's not the keyboard he was testing with.

MakaraSok:
> yields sil_khmer (includes brb-khmr)

`brb-khmr` was used to avoid the build complaint. Back then the compiler refused to accept `brb` in Developer (https://github.com/keymanapp/keyman/issues/4862). If now the Developer can build successfully with the script tag `khmr`, it can be removed.

> That's not the keyboard he was testing with.
@jahorton You are right. This model is built for `sil_brao`.

darcywong00:
> This repo will need the lexical model compiler update

You'll want to rebase when #132 gets merged

MakaraSok:
image

The .ts file turns out to be like this after updating to 14.0.272.

MakaraSok:
TODO:

- [ ] update the wordlist expected to be provided by the language community next week
- [ ] localize the metadata into KM as the target community barely knows English

MakaraSok:
@darcywong00 

I still have this issue of LM fail to give relevant suggestion when the cue is a cluster.

image

MakaraSok:
Two known issues need to be fixed in the future, but they shouldn't prevent this PR from moving forward.
Issues:
1. The wordlist is based on a working wordlist. Hope to have a complete wordlist in the future.
2. The model has some glitches because are using a temporary fix to mitigate the word break.

DavidLRowe:
@MakaraSok Are you waiting for a review by @darcywong00 ?

MakaraSok:
Can anyone look into this and approve so that the community can at least benefit from this? A fix may take a long while to realize.
@DavidLRowe @LornaSIL

DavidLRowe:
@MakaraSok I'll try to look at this tomorrow. I had assumed you were waiting on Darcy for a review.

DavidLRowe:
The .model_info file has `brb-khmr` but .kps file has both `brb` and `brb-Khmr`. (Also, .model_info has lowercase `k` in `khmr` rather than uppercase `K`, but I'm not sure if that is relevant.) At any rate, the entries in both files should agree, so probably you need to change the .model_info file to match what is in the .kps file.

Is there a problem with the .model.ts file? As you noted above, Developer says it cannot parse the file, apparently because of the replacement function for `wordBreaker`. 
```
  wordBreaker: function(str: string) {
    return str.split(/\s/).map(function(token) {
      return {
        left: str.indexOf(token),
        start: str.indexOf(token),
        right: str.indexOf(token) + token.length,
        end: str.indexOf(token) + token.length,
        text: token
      }
    });
  }
```
But maybe there's no problem with that code and it's just an issue with Developer not understanding it. If that's the case, I'm happy to approve this (once the BCP 47 tag issue is fixed), but I wanted to check. (I also note that there are no frequency counts in the word list, though I'm sure you are aware of that.)

release/sil/sil.brb.brao/sil.brb.brao.model_info
```diff
@@ -0,0 +1,7 @@
+{
+    "license": "mit",
+    "languages": [
+        "brb-khmr"
```
darcywong00, line 4: The kps file also uses `brb`. Do you want to add that here?

MakaraSok:
The language tag has been updated. Tested on Android and iOS simulators, it is working as expected.

PR #125, opened 2021-03-18
fix: Update LM compiler version and support Android 5.0

### LINKED ISSUE #124: [nrc.en.mtnt and katelem.ann-latn.getat] Rebuild required due to compiler bug
We will need to update the two models mentioned in keymanapp/keyman#4716 after we get the updated compiler.

_Originally posted by @mcdurdin in https://github.com/keymanapp/keyman/issues/4718#issuecomment-801682401_

Should be version 14.0.265-beta or later.

mcdurdin:
Fixes #124.

I haven't updated copyright dates because the change is inconsequential and actually only a version increment to get the new version deployed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,27 +5,20 @@
   "requires": true,
   "dependencies": {
     "@keymanapp/lexical-model-compiler": {
-      "version": "14.0.247-beta",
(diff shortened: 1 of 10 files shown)


DavidLRowe (approved):
welcome.htm for katelem.ann-latn.getat has "Lexica Model" that I think is probably intended as "Lexical Model", but I don't think it's worth changing.

PR #130, opened 2021-04-08
[cabecar lexical model]

victoriaq22:
updated 4/2021 and reuploaded to new fork

PR #134, opened 2021-05-12
[akevai_oh_zhou_coto_quint] CIM initial upload

victoriaq22:
Initial submission of Cook Islands Maori lexical model.  Sorry the name is so long, we had a lot of collaborators!  Thanks!


keyman-server:

DavidLRowe:
Looks pretty good. A few nit-picky details:

- The version number is 1.2 in HISTORY.md, 1.1 in README.md and 1.0 in the .kps file. It looks like the 1.1 entry in the README.md file should really be part of the HISTORY.md. Generally we just have a single version number entry in README.md that shows the latest version (with the details reserved for HISTORY.md), so that version could be set to 1.2. Then in the .kps (Packaging) file, on the Details tab, you can set the version number to 1.2.
- The copyright statement should probably be updated to 2020-2021 everywhere it appears.
- The Keyman team is in the process of changing how language designations are listed. In the past we specified `rar-Latn`. The newest versions of Developer warn that `rar-Latn` should really be `rar`. 
  - You should probably include both `rar` and `rar-Latn` in the list of languages on the Lexical Models tab. Eventually the keyboard will likely change to using `rar` rather than `rar-Latn` and you want the lexical model to connect with either one.
  - In the Projects menu, Project Settings item, untick the "Treat compiler hints and warnings as errors" box, so that the warning about `rar-Latn` won't prevent the model from building.
- It would be nice to have a better description field in the .model_info file.
- Do you want to add casing information? (See changes you made to the model.ts file in #130.)

BTW, there seem to be nine places where two forms of the same word occurred. The newest version of the compiler is smart enough to detect duplicates and add the frequency counts (whereas before we had to find and combine them manually). So the word list should be fine as is, but you might want to combine those yourself. (Compiling the model should give you a list of the words in question.)

victoriaq22:
I made the corrections suggested.  Thank you!

DavidLRowe (approved):

DavidLRowe:
Just for the record, to correct a comment I made earlier: 
With a more recent version of Developer (14.0.274.0, for example) it is no longer necessary to turn off the "Treat compiler hints and warnings as errors" option to allow a non-canonical BCP 47 code to be used.

You don't need to change anything in the project file.

mcdurdin:
This failed to deploy because the release/akevai_oh_zhou_coto_quint/akevai_oh_zhou_coto_quint.rar-latn.cim/akevainicholas_oh_zhou_cotosolano.rar-latn.cook_islands_maori.model_info file name does not match the folder name (akevai_oh_zhou_coto_quint.rar-latn.cim).

Note that overly long names may cause issues on some systems (Windows in particular).

Looking at the three parts of the filename:
* If you can find a way to shorten the 'author' part of the name (even an abbreviation e.g. `aozcq`), I would recommend doing that. This name is only for internal use anyway -- the information that gets displayed to end users comes from the content of the .model_info and the .kps files.
* The BCP 47 tag is fine
* Somewhat ironically, I think the model name itself would be better as `cook_islands_maori` than `cim`.

So overall, the name I recommend for both folder and filename is: `aozcq.rar-latn.cook_islands_maori`. This should be short enough to not be a problem on Windows. These names need to be consistent throughout all references.

PR #135, opened 2021-05-13
chore: drops deprecated includes

### LINKED ISSUE #133: chore: Cleanup includes/ folder?
As of stable-14.0, are the Type files in includes/ still needed?

LMLayer.d.ts
LMLayerWorker.d.ts
message.d.ts

If they're meant to keep in sync with the main repo as the readme says, I note they're out of date with common\predictive-text\build in the main repo.

Maybe cause we're using https://www.npmjs.com/package/@keymanapp/models-types now?

jahorton:
Fixes #133.

After some thought, I'm fairly certain the original intent for the 'includes' was to support custom models.  Note that we don't currently provide any of those.

That said, should someone wish to provide a custom model... it is my understanding that custom models should _not_ rely on these files anyway.  They should instead adhere to the specs from the `@keymanapp/models-types` module, which is already available as a dependency for the lexical model compiler.  This does not block access to our provided wordbreaking-functions, either.

So, I second the motion to eliminate these includes; they should not ever see use within a compiled model.  Should that need to change, we should probably examine what parts of the design require additional access and find a way to encapsulate the dependencies, rather than compiling against predictive-engine code.


darcywong00 (approved):
lgtm

mcdurdin (approved):

PR #138, opened 2021-05-23
Cook islands maori

victoriaq22:
Here are the updated files that David helped with. Thanks!

PR #139, opened 2021-05-27
remove renamed model

DavidLRowe:
The model in the akevai_oh_zhou_coto_quint folder failed to deploy because of a name mismatch. Marc recommended shortening the author name, which was done on #138 . This PR removes the old folder and its contents


darcywong00 (approved):
lgtm

Issue #112, opened 2021-02-19
Add `bal` to `bcc` lexical-models


darcywong00:
Yes, we should update those models to also include `bal` (don't remove `bcc` at this point to maintain compatibility with 13.0 keyboards)

_Originally posted by @mcdurdin in https://github.com/keymanapp/keyboards/pull/1452#discussion_r574763989_

DavidLRowe:
IIUC `bal` is a macrolanguage tag for Balochi and `bcc` is the tag for Southern Balochi. I don't know enough about this language situation to know whether that is going to cause problems, for example, someone from another Balochi language selecting `bal` and getting Southern Balochi predictive text. If @rmlockwood is available, it would be good to get his opinion.

darcywong00:
keyboards PR keymanapp/keyboards#1452 was abandoned, so this issue may not need to be resolved.

rmlockwood:
Just coming off of sabbatical. What do I need to do for this? Thanks.

DavidLRowe:
The issue is that 'bal' macrolanguage uses 'bcc' as its default, so 'bal' is preferred over 'bcc' as the canonical tag. 

It seems that keyboards and lexical models that currently have 'bcc' should at least include 'bal', and perhaps should include both 'bcc' and 'bal'.  @mcdurdin Can you confirm?

mcdurdin:
At present, lexical models should list both `bcc` and `bal` as the matching for BCP 47 here does not do this level of canonicalization. Keyboards should only include one.

DavidLRowe:
The lexical model for bcc-arab includes entries for bal, so that should be fine.

The sil.bcc-latn.upp_ptwl1 lists "bcc-Latn", but not "bal-Latn", in *.kps and *.model_info files.
Ideally, this should be changed to include "bal-Latn" (but keep "bcc-Latn" for backward compatibility).

Two keyboards for Latin-script Balochi need to switch from bcc-Latn to bal-Latn (see https://github.com/keymanapp/keyboards/issues/1593), but after the change to the lexical model.

Note also: https://github.com/keymanapp/lexical-models/issues/119

PR #145, opened 2021-07-14
added language code bal

rmlockwood:


keyman-server:

DavidLRowe:
I see that your *.ts file is more complex than the default one, so perhaps issue https://github.com/keymanapp/lexical-models/issues/119 isn't relevant. (If that's the case, you can make a comment to that effect and close the issue.)

If the casing issue is relevant, would you want to include that change in this PR?

rmlockwood:
Ok. I added the UseCasing.

DavidLRowe (approved):

DavidLRowe:
fixes #119 
fixes #112

PR #146, opened 2021-07-19
GFF Gurage Lexical Model v1.1

dyacob:
This update remaps applicable characters into the Unicode 14 Ethiopic Extended-B block for compatibility with the GFF Gurage Keyboard updates.


keyman-server:

DavidLRowe (approved):

PR #147, opened 2021-07-24
Create English (Shavian alphabet) ReadLex lexical model

Shavian-info:


keyman-server:

DavidLRowe:
In order to fit the naming convention (which keeps our automated build system happy), we need to have a consistent "author.en-shaw.readlex" name (used for the project name and the base name of many of the files in the project), where the "author" part is also used as the folder in which the project folder resides. So, for example, instead of 
```
shavian\en-shaw.readlex\shavian_info.en-shaw.readlex.kpj
```
we would need
```
shavian_info\shavian_info.en-shaw.readlex\shavian_info.en-shaw.readlex.kpj
```
and so on for all the files (assuming you choose `shavian_info` as the author name).

Sorry to be so picky, but it makes maintaining the build system easier. 

If it's too much trouble to rename the files in this pull request, please feel free to close this PR and open a new one.

Thanks for your contribution!

Shavian-info:
I've tried to correct the folder names to match the required naming conventions.

DavidLRowe:
@Shavian-info Looks good! The only question I have is about the source/english_shavian.jpg file. It is included in the package, but not specified as the image file (on the Details tab), so I don't think anything will use it. (Actually, I'm not sure I've seen any lexical model packages that have included a custom image, so I don't know whether that is currently implemented or not.)

I'd suggest that you either specify that .jpg file as the image file in the package source, or else remove the .jpg file from the source folder (and from the .kps package source).

Thanks!

Shavian-info:
I've added in the reference to the Graphic File. Even if not currently supported, happy to keep it if it doesn't cause any issues.

DavidLRowe (approved):

PR #148, opened 2021-11-05
[mon_wordlist] New submission on behalf of author

LornaSIL:
Relates to new keyboard submission.


DavidLRowe:
The README.md description has "Cook Islands Maori" (Also readme.htm, welcome.htm). It would be good to change those to reference "Mon".

@darcywong00 or @mcdurdin What is the current advice for naming lexical model projects? According to langtags.json, 
"mnw", "mnw-mymr", and "mnw-mymr-mm" are all equivalent. The associated keyboard is using "mnw". This project is currently using ait.mnw-mymr-mm.mon . Should it be ait.mnw.mon instead?

mcdurdin:
> Should it be ait.mnw.mon instead?

Yes, I would go with the minimal tag.

DavidLRowe:
@LornaSIL Unfortunately, that means changing all the file and folder names to use `ait.mnw.mon` instead of `ait.mnw-mymr-mm.mon`.

DavidLRowe (approved):

PR #151, opened 2022-01-05
[sil_tchad_dzg] Initial submission of Dazaga predictive keyboard

jeffheath-sil:
@DavidLRowe First of two predictive keyboards that we would like in the Keyman repo (for download from the Internet)


keyman-server:

DavidLRowe:
Hi @jeffheath-sil. 

In order to fit the scheme of the automated build system, your `sil_tchad.dzg.lexique` folder needs to be in a folder named `sil_tchad` (rather than `sil`). You can move the files to the new folder or (which may be easier) close this PR and start a new one, 

I do note there were a number of warnings about duplicated words, but the compiler now can handle that and sums the counts for the entries.

Everything else looks fine. I don't know if the readme.htm and welcome.htm are currently used, but it's fine to have them there.

jeffheath-sil:
Moving this PR to the sil_tchad folder

PR #152, opened 2022-01-06
[sil_tchad_dzg] Initial submission of Dazaga predictive keyboard

jeffheath-sil:
@DavidLRowe You suggested putting the lexical model in the sil_tchad folder, to accomadate the automated build, but I had lots of problems doing that. So I went back to the project and renamed it to start with "sil.", so it would be the same as the other SIL models. I also removed some of the warnings (duplicated words), but left the ones in where they had actual counts (so there are still some warnings, but not as many).


keyman-server:

DavidLRowe (approved):

PR #153, opened 2022-01-06
[sil_tchad_tuq] Initial submission of Tedaga predictive keyboard

jeffheath-sil:
@DavidLRowe Here is the second lexical model, modified (sil_tchad -> sil) the same way as the other model.


keyman-server:

DavidLRowe (approved):

PR #155, opened 2022-02-18
use composed as the end result

rmlockwood:
This is basically the same change I did for bcc-Latn a couple of months ago.


keyman-server:

DavidLRowe (approved):

PR #158, opened 2022-02-26
Added the Postmodern English US Midland lexical model

PostmodernEnglish:
This is a lexical model for Postmodern English when it is written in a US Midland accent. The way words are written in Postmodern English will vary depending on the accent of the person writing.

For the wordlist counts, we went with the following strategy for now:
-If the word is in the top 200 English words, add 20
-If the word is a root word, add 10

We chose to use the en BCP 47 tag for the moment, since no specific tag currently exists for Postmodern English. Different lexical models for different accents will be created in the future. Please let us know if the directory or file names need to be adjusted.


DavidLRowe:
@PostmodernEnglish Ping me when you're ready for a review.

I would suggest that you remove the copyright line from README.md. We no longer require the copyright line in as many files as we did in the past, and it means fewer files to update for a new version. (Alternatively, you could keep the copyright statement, but omit the date. That's acceptable everywhere but the LICENSE.md file.)

PostmodernEnglish:
That sounds like a plan. We've identified a number of mistakes after committing, so we've decided to go over the list once more, which could take some time. We'll let you know when we've finished, of course.

darcywong00:
>We chose to use the en BCP 47 tag for the moment, since no specific tag currently exists for Postmodern English. 

I think our team will still need to have discussions on how to handle this. We'll want to associate your Postmodern English dictionary with your [postmodern_english_us_natural](https://keyman.com/keyboards/postmodern_english_us_natural) keyboard, and not have it offer suggestions for standard English keyboards.

PostmodernEnglish:
@DavidLRowe We're ready for this to be reviewed again. Please let us know what your team decided to do about the BCP 47 tags. Thanks.

DavidLRowe:
@darcywong00 Do you have a solution for the BCP 47 tag question?

darcywong00:
> @darcywong00 Do you have a solution for the BCP 47 tag question?

I've asked @mhosken. Skimming through the wordlist, I notice the Script also uses a few non-Latin characters:

ч U+‎0447 CYRILLIC SMALL LETTER CHE
ԅ U+0505 CYRILLIC SMALL LETTER KOMI ZJE

darcywong00:
Ideally, we would use a private-use variant like `en-x-postmodern-midland` but Keyman currently doesn't handle private-use tags.

Maybe we can use a private-use region (QM..QZ) like `en-QM`?

Will the lexical-model compile if you use that language tag?

release/pme/pme.en.postmodern_english_us_midland/source/pme.en.postmodern_english_us_midland.model.kps
```diff
@@ -0,0 +1,55 @@
...
+      <FileType>.js</FileType>
+    </File>
+  </Files>
+  <Keyboards/>
+  <LexicalModels>
+    <LexicalModel>
+      <Name></Name>
+      <ID>pme.en.postmodern_english_us_midland</ID>
+      <Languages>
+        <Language ID="en">Postmodern English</Language>
```
darcywong00, line 50:
```suggestion
        <Language ID="en-qm">Postmodern English (Midland)</Language>
```

darcywong00 (dismissed):
> We chose to use the en BCP 47 tag for the moment, since no specific tag currently exists for Postmodern English. Different lexical models for different accents will be created in the future. Please let us know if the directory or file names need to be adjusted.

Since Keyman currently doesn't support private use variants, can you try using a private use Region? e.g. `en-QM` instead of `en`. This would affect all the filenames

For your future lexical-models, you could try other private use regions (`QM`..`QZ`)

You'll also need to update your keyboards to match the lexical model language tags.

release/pme/pme.en.postmodern_english_us_midland/source/pme.en.postmodern_english_us_midland.model.kps
DavidLRowe: @darcywong00 Typically region qualifiers are all caps (-QM rather than -qm). Would it be preferable to use "en-QM" here?

darcywong00: On this line, "en-QM" would work.

The filenames elsewhere will need to use lower-case `en-qm`

DavidLRowe:
@darcywong00 Thanks for that clarification. I'll try to summarize. Please correct me if I've misunderstood.

@PostmodernEnglish In summary, the project file and folder names will use
`pme.en-qm.postmodern_english_us_midland`
The .model_info file and the .kps file will use "en-QM" within the file itself.
You'll need to update your keyboards to support "en-QM" in the .keyboard_info file and the .kps file. If you add another lexical model (hypothetically for the Northeast) you could pick another Private Use region code (QN for example), and add "en-QN" to the keyboards. You'll want to explicitly spell out what region you are targeting with your Private Use region codes: en-QM English (US Midland).

Hopefully that's enough information for you to move ahead, but if not keep asking questions!

PostmodernEnglish:
@DavidLRowe  Any specific Postmodern keyboard layout is not 1:1 with any specific written accent. For example, for a US Midland accent style of writing, we could use any Postmodern keyboard to accomplish this (e.g. a US Natural or a UK DualStroke keyboard). It's also important to note that some users of Postmodern will use more than one accent style in a mixed fashion- such as if someone lives close to the boundary between US Midland and US Southern. With these things in mind, it seems like there are two ways to go about this:

1.) Every accent variant has its own BCP 47 tag, and all the keyboards will be tagged with all the BCP 47 tags that exist for Postmodern.
2.) All accent variants have the same BCP 47 tag, such as en-QP, and all keyboards are only tagged with this single tag.

To our understanding, both of these methods would accomplish the same thing. Let us know if it would still be preferable to use a different custom BCP 47 tag for each Postmodern accent. Thanks.

DavidLRowe:
I was initially thinking of your scenario (1), but I think your scenario (2) has a lot of merit. You'll have the distinction of pme_lexical_us_midland vs pme_lexical_uk_north (hypothetical example) in choosing a lexical model that fits the dialect, and all can use the `en-QP` tag (as will all the keyboards). This also reduces the load of adding a new lexical model, since you would no longer have to add the new tag to all the keyboards as well.

PostmodernEnglish:
We decided on using a universal BCP 47 tag of en-QP, which all keyboards and lexical models will use. The user will have to pick and choose which Postmodern lexical models they want to use. If everything looks okay, feel free to approve and merge. We'll also update our keyboards to match this new BCP 47 tag fairly soon. Thanks.

PR #159, opened 2022-03-07
Two test dictionaries -- Tlicho and Chipewyan

ind-nt:
Work when packaged into kmps and downloaded through Keyman app.


DavidLRowe:
In the iles.chp.indigenous_nt/source/wordlist.tsv file, the first uncommented line is "word count". I suspect that should be deleted or made into a comment line. As it is, it suggests "word" as a possibility.

The iles.dgr.indigenous_nt/source/wordlist.tsv file has same issue.

Both files contain decomposed characters (base character plus separate combining diacritic), which have been converted to composed form. This is not a problem (as it was in the initial days), but you should be aware of it. (This could potentially cause confusion if the keyboard produced decomposed forms except when a word suggested by the predictive text model was selected, but that doesn't seem to be the case here.)

So, if you could comment out those lines, this should be ready for approval. Thanks!

ind-nt:
Thanks David. That's great. I've commented out the headers for both files.

DavidLRowe:
@ind-nt Thanks for those fixes to the lexical models. Now there are some changes needed to the package source (.kps) files to allow our automated build system to correctly package these models.

In the package for iles.dgr.indigenous_nt, the path to the .js file should point to the build folder rather than the source folder (that is, to `iles\iles.dgr.indigenous_nt\build\iles.dgr.indigenous_nt.model.js` rather than `iles\iles.dgr.indigenous_nt\source\iles.dgr.indigenous_nt.model.js`). In Developer, it's easiest to remove the existing .js entry and then add and navigate to the correct one. You also need to specify "dgr" as a BCP 47 tag in the Lexical Models tab.

In the package for iles.chp.indigenous_nt, the .kps file fails to validate, I guess because no files are present in the list of files. You'll need to add the .js file from the build folder, and the readme.htm and welcome.htm files from the source folder. In addition, you'll need to add the "chp" BCP 47 tag and fill in the information on the Details tab.

Let me know if you have any questions.

ind-nt:
Hi David,

Ok, I corrected the path for the .js file. The BCP 47 tags do appear in the kps files, for chp, the readme and welcome are specified. I compared with other packages and it seems ok.

- R

DavidLRowe:
The .kps file (in the source folder) is misnamed. You have `ile.chp.indigenous_nt.model.kps`, but it needs to be `iles.chp.indigenous_nt.model.kps`.

ind-nt:
Ok, corrected, not sure how that happened!

DavidLRowe (approved):

PR #167, opened 2022-06-04
Added words for the Postmodern English lexical update wave 2

PostmodernEnglish:
Added new words to the US Midland accent, for the wave 2 update.

### DIFF
--- a/release/pme/pme.en-qp.postmodern_english_us_midland/HISTORY.md
+++ b/release/pme/pme.en-qp.postmodern_english_us_midland/HISTORY.md
@@ -8,3 +8,7 @@ v1.0.0 (2022-02-26)
 v1.0.1 (2022-04-10)
 ----------------
 - Wave 1 update of the lexical model.
+
+v1.0.2 (2022-06-04)
+----------------
+- Wave 2 update of the lexical model.
--- a/release/pme/pme.en-qp.postmodern_english_us_midland/pme.en-qp.postmodern_english_us_midland.kpj
+++ b/release/pme/pme.en-qp.postmodern_english_us_midland/pme.en-qp.postmodern_english_us_midland.kpj
@@ -40,12 +40,12 @@
       <ID>id_f3f4c2984a0f3c656b7bb13b630ab777</ID>
       <Filename>pme.en-qp.postmodern_english_us_midland.model.kps</Filename>
       <Filepath>source\pme.en-qp.postmodern_english_us_midland.model.kps</Filepath>
-      <FileVersion>1.0.1</FileVersion>
+      <FileVersion>1.0.2</FileVersion>
       <FileType>.kps</FileType>
       <Details>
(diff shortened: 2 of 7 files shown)


PostmodernEnglish:
Question- do we need to also update the keyboard versions of our keyboards for them to recognize these changes, or are the version number changes in this lexical model sufficient?

keyman-server:

DavidLRowe:
The keyboard should not need to be updated. The keyboard-model link is based on the language tag (en-QP in this case). IIUC there is a setting for Keyman to check for updates to the model and, if an update is found, to offer to download the updated model.

PostmodernEnglish:
@DavidLRowe It seems that our keyboards are not recognizing this lexical model update. Even if we try to remove and re-download one of our keyboards, it still reports that the dictionary version is only 1.0.0 (this merge is 2 versions after that, being 1.0.2). We're not sure if this is something we're doing wrong, or there is a bug happening elsewhere in Keyman. Please look into this, as our lexical updates do not seem to be getting pushed out. Thank you.

DavidLRowe:
Local testing using emulator in Keyman Developer:
From Keyman Developer (v 15.0.256.0), loaded the lexical model, Build tab, "Test Lexical Model in Web Browser".
Browse to select postmodern_english_us_natural.js from build folder for postmodern_english_us_natural keyboard (which I had previously compiled).
"Test Lexical Model" button, then "Open in browser" button.
For Keyboard dropdown list, selected "postmodern_english_us_natural".
For Model, selected "pme.en-qp.postmodern_english_us_midland".
For Device, selected "iPad Mini"

Typing "alī" shows "alīŋ" as the first proposed word. 
Typing "ҁūd" shows "ҁūdn't" as the first proposed word.
Typing "dīsé" shows "Dīsémbԅ" as the first proposed word.
"alīŋ" and "ҁūdn't" and "Dīsémbԅ" were added in the 1.0.2 revision.

So the model (v1.0.2) works with the keyboard in the emulator.

I then tried installing the postmodern_english_us_natural keyboard on an Android phone, and it seems to have installed pme.en-qp.postmodern_english_us_midland lexical model v1.0.1 automatically. I can uninstall that dictionary, but can't seem to find v1.0.2. 

At this point, I'm out of my depth, so will ask the Keyman developers for help!

DavidLRowe:
Turns out there is a problem on the Keyman site. @mcdurdin is investigating.

mcdurdin:
I have manually pushed a release out for this model, so it should be online now.

DavidLRowe:
When testing on an Android phone, I uninstalled and reinstalled the dictionary but couldn't get it to move from v1.0.1 to v1.0.2 (even after the manual update). Uninstalling the keyboard and reinstalling it automatically installed the v1.0.2 dictionary for me just now, but I don't think reinstalling the keyboard is intended to be necessary. 

@PostmodernEnglish Sorry you ran into this issue. Thanks for notifying us and keep us posted.

PR #172, opened 2022-07-07
Obolo LM 1.1

katelem24:


keyman-server:

DavidLRowe:
Now you need to add back in the changed files (wordlist, etc.).

katelem24:
> Now you need to add back in the changed files (wordlist, etc.).

Done. Please cross-check.

DavidLRowe:
The list of files in the package source (.kps) file includes ann_kb_48.ico which does not exist in this project. You either need to delete this name from the list of files, or else include it in the source folder (and change the location information in the list of files).

katelem24:
Done

DavidLRowe (approved):

PR #173, opened 2022-07-14
[dotland_ru_russian] Lexical model for the Russian language.

dotland:
This model includes more than 250 000 Russian words with usage frequencies. Those were analyzed from the texts of classical literature by PrimerPrep utility.


keyman-server:

PR #174, opened 2022-07-25
[dotland.hy.armenian] Initial submission of armenian lexical model

dotland:
The initial submission of the Armenian lexical model which contains about 160 000 words.

Links: [https://github.com/dotland/lexical-model-hy](https://github.com/dotland/lexical-model-hy)


keyman-server:

dotland:
Reduced the wordlist to 85 000 removing rare words, which improves performance.
Add language casing for capitalized text suggestion.

DavidLRowe (approved):

Issue #143, opened 2021-07-01
[nrc.en.mtnt] More contractions in suggestions


MakaraSok:
A user suggested that more contractions are given in the suggestion banner of the EuroLatin (SIL) keyboard.

darcywong00:
The lexical-model we're using for English is nrc.en.mtnt/ and the wordlist is mtnt.tsv

I do note the contractions in the wordlist use "left single quotation mark" `‘` instead of "apostrophe" `'`

DavidLRowe:
I agree that U+2018 "left single quotation mark" is not the right character. 

If the lexical model uses U+0027 "apostrophe", what will happen when this model is used with a keyboard that changes apostrophe (U+0027) to "right single quotation mark" (U+2019). If I type "did" and then pick "didn't" from the list, it seems as though I'll get the U+0027 from the .tsv file rather than the U+2019 that the keyboard would generate. OTOH if I type "didn'" and the keyboard changes the U+0027 to U+2019, it seems it will no longer offer me "didn't" as a possibility because the apostrophe no longer matches. 

But there's lots about this process that I don't understand!

darcywong00:
With a context of "were", I get 3 suggestions:
"were" (frequency 5309)
"weren't" (frequency 385)
"werent" (frequency 16)

"we're" has a frequency of 927 so I would expect it to appear

I wonder if we should also scrub the following (typos and symbols) from the wordlist:
werent	16
hasnt   23
wouldnt 80
couldnt 37
shouldnt  39
dont 977
hes  116
arent 73

$	1898
√	(square root, frequency 19)
______	6
▢	9
ಠ_ಠ	9

DavidLRowe:
@darcywong00 I agree that there are typos and symbols that should be removed.

What about proper names (Danae, Kristin, Derrick, etc.)? Or Catan?!

What about 32GB and 6g?

The question is who would like to undertake this effort?

jahorton:
> I agree that U+2018 "left single quotation mark" is not the right character.
> 
[...]

PR #177, opened 2022-08-24
fix: [nrc.en.mtnt] Replace left single quote with apostrophe

darcywong00:
Addresses the TODO of #143 in replacing the wordlist use of "left-single-quote" with "apostrophe".

However, this doesn't seem to improve the suggestions of contractions .

When I installed with Keyman for Android 16.0, typing "ill", "hes", etc didn't offer contractions as suggestions...


darcywong00:
From standup meeting, we'll hold on this PR until @jahorton is able to implement the fix (https://github.com/keymanapp/lexical-models/issues/143#issuecomment-1096934795 ?) at a future sprint...

jahorton:
Hmm.  Granted, all of the following tests were done on the Web-only test page, but...

image

But...

image

When extra-precise:

image

But being a little too close to 'j' with the first keystroke can prevent `he's` from showing as a suggestion - like, it wouldn't even be in the top 12.

image

vs

image

Still shows up.

jahorton:
Granted, there _is_ the issue that we English speakers "expect" apostrophe omission (for contractions) to count as "less" of an error than forgetting a standard letter... but that's language-specific knowledge, and thus might be a bit harder to model properly.

jahorton:
To be clear, this is using the build from https://github.com/keymanapp/keyman/pull/7612, which does not target this issue in any way.

jahorton:
> From standup meeting, we'll hold on this PR until @jahorton is able to implement the fix ([#143 (comment)](https://github.com/keymanapp/lexical-models/issues/143#issuecomment-1096934795) ?) at a future sprint...

This tidbit is a separate issue.  That's worth addressing to be sure, but is not sufficient reason to hold back this PR.

jahorton (approved):

PR #180, opened 2022-12-11
Added Updates for Wave 3 of the PME Lexical Model

PostmodernEnglish:
Added new words to the US Midland accent, for the wave 3 update.

### DIFF
--- a/release/pme/pme.en-qp.postmodern_english_us_midland/HISTORY.md
+++ b/release/pme/pme.en-qp.postmodern_english_us_midland/HISTORY.md
@@ -12,3 +12,7 @@ v1.0.1 (2022-04-10)
 v1.0.2 (2022-06-04)
 ----------------
 - Wave 2 update of the lexical model.
+
+v1.0.3 (2022-12-10)
+----------------
+- Wave 3 update of the lexical model.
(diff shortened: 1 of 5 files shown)


keyman-server:

PostmodernEnglish:
We are currently in the process of doing a final overview of the changes to look for errors. We'll let you know when we're ready for a merge. Thank you.

DavidLRowe:
Realistically it will be January before someone can review this, so take your time!

PostmodernEnglish:
We're now ready for your team to review whenever. Thank you.

PostmodernEnglish:
We're now ready for your team to review whenever. Thank you.

DavidLRowe (approved):

DavidLRowe:
@PostmodernEnglish Thanks for your patience!

PostmodernEnglish:
@DavidLRowe 
There appears to be a problem with this build of the mobile keyboard (and previous builds for our other keyboards). We've noticed that whenever we attempt to select either the "emoji" or "number pad" keyboard layer in any of our 4 keyboards, Keyman reports an error, and the user is unable to get to these keyboard layers. The other 3 standard keyboard layers appear to work fine, however. The error Keyman reports when this happens is something like "Error for the keyboard postmodern_english_us_natural for the language .". It seems that something about the lexical model might be causing trouble for Keyman. Could it be related to the fact that our lexical model uses the BCP 47 code "en-qp"? We've verified this on two different Android devices running the latest version of Keyman. Clearing the app cache and restarting the devices didn't change the outcome, nor did reinstalling the app.

Whenever your team is able, please look into this. Thank you.

LornaSIL:
Your `Núm` and Emoji buttons don't have an `ID` in at least the postmodern_english_uk_natural keyboard. I suspect it's related to that. However, I don't know what the ID should be... I would expect `K_NUMERALS` and `K_SYMBOLS` to be acceptable.

image

PR #183, opened 2023-01-23
sil_senegal.bsc.wl1

madskinner2:
Production release of sil_senegal.bsc.wl1


keyman-server:

DavidLRowe:
Normally, I'd expect a date range 2021-2023 in the copyright statement, but I guess it's fine to use 2021 for this initial version, then update it later (for example, with a revised word list).

sil_senegal.bsc.wl1.model_info
- please change the description to something that will help people identify this lexical model, perhaps (for example): "predictive text model for Oniyan (bsc, Senegal), based on a word list"

README.md
- Take the description you came up with for the .model_info file and use it here also. 
- Consider removing the version number line, so you don't need to change this file when updating the lexical model.
- Consider removing the copyright statement (or at least the date), so you don't need to change this file when updating the lexical model.

readme.htm
- Take the description you came up with for the .model_info file and use it here also. 
- Consider removing the copyright statement (or at least the date), so you don't need to change this file when updating the lexical model.

welcome.htm
- Take the description you came up with for the .model_info file and use it here also. 
- Consider removing the copyright statement (or at least the date), so you don't need to change this file when updating the lexical model.

madskinner2:
updated documentation

DavidLRowe (approved):

PR #184, opened 2023-01-27
initial cou, dyo, gsl, gsl-ajami, krx, ndv, sav, srr and wo upload

madskinner2:


keyman-server:

DavidLRowe:
Generally the copyright date range should match the information in the HISTORY.md file, just to avoid questions, but everything else seems okay, so you can correct it when you update these lexical models.

DavidLRowe (approved):

PR #185, opened 2023-02-01
chore: Update lexical-model compiler to 16.0.138

darcywong00:
Accompanying the Keyman 16.0.138 release, this updates the lexical-model compiler from 15.0 to 16.0

There's a multititude of compiler warnings
* duplicate words found in a (wordlist) file; summing counts
* entries not being Unicode NFC. Automatically converting to NFC

But the overall build passes

I note node is still  `"^14.15.0",` which we should update to LTS at some point.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,8 +9,8 @@
       "version": "1.0.0",
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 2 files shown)


darcywong00:
We'll update Node on a separate PR

mcdurdin (approved):

PR #196, opened 2023-04-12
Himyarit Musnad Lexical Model v1.0

dyacob:
This pull request is made on behalf of the author, @RiyadhAlfareh . It includes a simple vocabulary to assist in typing of text with the Himyarit Musnad keyboard.


RiyadhAlfareh:
Thanks Daniel for your support

في الأربعاء، 12 أبريل 2023 3:22 ص Daniel Yacob ***@***.***>
كتب:

keyman-server:

DavidLRowe:
When I open the project with Keyman Developer and look at the package (.kps), selecting the "Lexical Models" tab on the left, I see that the "Is Right-to-left" box is not ticked, though I think this is a RTL script. I'm not sure I've run into this before. It seems that you should probably make that change.

@mcdurdin How critical is it to have the RTL box ticked in the Lexical Models tab? (Just curious!)

dyacob:
Thanks @DavidLRowe , I'm a tad surprised because I was very sure that I had checked that box ... and I believe we have uncovered a bug!   My Keyman Developer version is 16.0.139, these steps bring out a bug where the "Is Right-to-left" will become _unchecked_ : 

1. Open a lexical model .kps file.
2. Go to the "Lexical Models" section/tab.
3. Check "Is Right-to-left"
4. Go to the "Source" section/tab (do nothing).
5. Return to "Lexical Models" -"Is Right-to-left" is unchecked!

"Is Right-to-left" remains checked if you go back and forth to any of the other tabs.  I'll open a bug report for KMD.

I'll make a new commit with "Is Right-to-left" checked, I guess it will be fine so long as the "Source" section is not visited.

DavidLRowe (approved):

PR #197, opened 2023-05-01
[gff.byn.gff_blin] Initial release

dyacob:
This update provides a cleanup of identical words that had separate entries due to punctuation use (apostrophe vs single left or right curly quote).


keyman-server:

darcywong00:
While you're cleaning up punctuation, I ran your wordlist through unist and saw 21 instances of period `.`. Just checking those are correct

dyacob:
Thanks for reviewing @darcywong00 .  The words with periods are regular abbreviations that I think most users should recognize.  The words with Latin script, like "ንVOA", are references to external or international organizations.  Though not native Tigrinya words, but as the frequency counts indicate they get used often in Tigrinya text.  What is the best policy to follow here regarding inclusion/exclusion?

darcywong00:
I guess it's ok to include a few. In general, I think we'd prefer the words match the specified language tag (`byn`).

release/gff/gff.byn.gff_blin/source/welcome.htm
```diff
@@ -0,0 +1,28 @@
...
+    h2 { font: bold 12pt Tahoma; color: #4444cc; }
+  </style>
+</head>
+<body>
+
+<h1>Start Using GFF Amharic</h1>
+
+<p>
+    This lexical model is intended for use with the Blin Keyman Keyboard on mobile devices, see:
+    <a href="https://keyman.com/blin/">https://keyman.com/blin/</a>
```
darcywong00, line 18: Is this now?
```suggestion
    <a href="https://keyman.com/keyboards/gff_blin">https://keyman.com/keyboards/gff_blin</a>
```

dyacob:
@darcywong00 a thought occurred to me on a potential problem.  The `gff_blin` keyboard uses the language ID `byn-Ethi` to distinguish itself from the `byn-Latn` convention (aka "Xaleget").  The lexicon uses the more general language ID `byn`; should this be changed to `byn-Ethi` as well in order for the keyboard to successfully find and download the lexicon?

darcywong00:
> @darcywong00 a thought occurred to me on a potential problem. The `gff_blin` keyboard uses the language ID `byn-Ethi` to distinguish itself from the `byn-Latn` convention (aka "Xaleget"). The lexicon uses the more general language ID `byn`; should this be changed to `byn-Ethi` as well in order for the keyboard to successfully find and download the lexicon?

Yes, the mobile apps will expect an exact match of the language ID to associate the lexical-model with a keyboard. (casing doesn't matter).

dyacob:
Thanks for confirming @darcywong00 , I've made the edit to sync the language ID.

DavidLRowe:
@darcywong00 Does the .model_info file need to have the same language ID change?

release/gff/gff.byn.gff_blin/source/gff.byn.gff_blin.model.kps
```diff
@@ -0,0 +1,54 @@
...
+      <FileType>.htm</FileType>
+    </File>
+  </Files>
+  <Keyboards/>
+  <LexicalModels>
+    <LexicalModel>
+      <Name>GFF Blin</Name>
+      <ID>gff.byn.gff_blin</ID>
+      <Languages>
+        <Language ID="byn-Ethi">Blin</Language>
```
darcywong00, line 49: Was this populated by Keyman Developer? We may want to add the script info like
```suggestion
        <Language ID="byn-Ethi">Blin (Ethiopic)</Language>
```


As @DavidLRowe pointed out, the .model_info should also have `byn-Ethi`.

release/gff/gff.byn.gff_blin/source/welcome.htm
DavidLRowe: @dyacob Did you want to make this change? The original link doesn't work, so needs to be changed or removed.

DavidLRowe:
@darcywong00 If the BCP 47 code changed to byn-Ethi, should the name of this model change to gff.byn-Ethi.gff_blin? I'm not totally clear on what is used to establish the connection between keyboard and model. Is it the BCP 47 code in the package? or does the middle part of the model name play a role. Or, turning the question around, is the BCP 47 code in the model name a convention that doesn't have to match exactly the code in the package? (I'll also note that in this case byn-Ethi and byn are equivalent tags.)

dyacob:
@DavidLRowe , thanks for alerting me to the update needed, I thought I had gotten in a week earlier but it must have been in the wrong branch.

DavidLRowe (approved):

dyacob:
@DavidLRowe & @darcywong00  -I'm very glad to see this PR merged and close.  I do have a troubleshooting question, I can post it elsewhere if you prefer.  I went into the Android Keyman app to install the Blin (byn-Ethi) dictionary, but it doesn't appear.  My steps:

1. Keyman > Installed languages > Bilin > Dictionary (Check for available dictionary)
2. The messages appear: "Checking for associated...." , "Downloading dictionary..." , "Dictionary download is finished" , "Resources successfully updated!"
3. The "Check for available dictionary" text remains, it is not replaced with the dictionary name.
4. Going into the Keyman app editor, or in local editors, and launching the Blin keyboard, the dictionary is not loaded, the top of the screen with the word options does not appear.
5. I've restarted Keyman, and restarted my phone, but no change. Other keyboards work fine.
6. Steps (1) and (2) are repeated, and the results are identical.

So I'm wondering if there is a bad metadata, or misaligned metadata, between the keyboard and dictionary.  It is interesting that the app can find, retrieve and install the dictionary, but it never becomes available when the keyboard is in use.

### Some version info:
* GFF Blin Keyboard v1.5.1
* Android Keyman: Version: 16.0.138

thanks!

DavidLRowe:
Thanks for the good report. It might be better to raise an issue at https://github.com/keymanapp/keyman/issues

PR #205, opened 2023-05-22
fix: updates example.en.custom

jahorton:
We hadn't exactly maintained the custom model for a while, so this should update it to working order.  I even tested it in KMW by temporarily linking it with one of the testing pages:

image

The scary thing is... it actually _did_ compile before this, despite some of the referenced types not actually existing.  I went ahead and made sure that all types currently referenced do exist.

### DIFF
--- a/sample/example/example.en.custom_word_breaker/source/example.en.custom_word_breaker.model.ts
+++ b/sample/example/example.en.custom_word_breaker/source/example.en.custom_word_breaker.model.ts
@@ -1,4 +1,5 @@
-/// <reference types="@keymanapp/lexical-model-types" />
+// Defines LexicalModelSource.
+/// <reference types="@keymanapp/lexical-model-compiler/dist/kmlmc.d.ts" />
 
 const source: LexicalModelSource = {
   format: 'trie-1.0',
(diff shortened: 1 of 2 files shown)


sample/example/example.en.custom_word_breaker/source/example.en.custom_word_breaker.model.ts
```diff
@@ -1,4 +1,5 @@
-/// <reference types="@keymanapp/lexical-model-types" />
+// Defines LexicalModelSource.
+/// <reference types="@keymanapp/lexical-model-compiler/dist/kmlmc.d.ts" />
```
jahorton, line 2: This allows VS Code Intellisense to link to the `LexicalModelSource` type.  Granted, I have a feeling that it's probably relocated a bit for 17.0... but I think the compiler almost always ignores the definition or lack thereof for the type.  It's mostly there for Intellisense when someone actually looks at the example.

mcdurdin (approved):
LGTM!

mcdurdin:
@eddieantonio you might also like to see this -- still hoping to see some n-grams or FSTs ~~or ChatGPTs~~ implemented against these interfaces!

PR #212, opened 2023-05-30
GFF Tigrinya Lexicon v1.2.

dyacob:
This update removes all words with the letter "ጛ" which comes from Blin documents that UniLex misidentified as Tigrinya.


keyman-server:

DavidLRowe (approved):

PR #216, opened 2023-07-04
GFF Tigrinya-Eritrea Lexical Model v1.0

dyacob:
The lexical model is based on the content found from 131 issues of the Haddas Eritrea (ሓዳስ ኤርትራ) newspaper.  The wordlist better reflects the spelling conventions, and word choices, of Tigrinya as spoken in Eritrea as compared to the predecessor "GFF Tigrinya" lexical model which in turn is based on the UniLex project which makes no regional distinction.


DavidLRowe:
The .model_info file lists "ti-ET" and "ti-ER" whereas .kps file lists only "ti-ER". It seems that the information in the two files should agree. From the comments it seems that "ti-ET" should be removed from the .model_info file.

dyacob:
@DavidLRowe -thanks for catching this!  A fix has been applied.

keyman-server:

DavidLRowe (approved):

PR #219, opened 2023-07-08
GFF Tigrinya-Ethiopia Lexical Model v1.0.1

dyacob:
This update fixes some minor but important issues:

* The dictionary display name (&lt;LexicalModel&gt;&lt;Name&gt;) is now populated.
* The `.model_info` file deprecates the region-neutral Tigrinya lexical model.
* `readme.htm` is in synch with the `README.md`
* `welcome.htm` is enhanced.
* Some resource versions (1.2) which were held over from the region-neutral `.kpj` file have been set properly.

Note that the wordlist is unchanged.  The `.kmp` file has been manually installed on an Android phone and verified.


keyman-server:

DavidLRowe (approved):

PR #222, opened 2023-09-20
[tk_arab] Lexical model for tk-arab

LornaSIL:
New lexical model. It probably needs a good check since I haven't done much with lexical models and what I knew, I've forgotten.


DavidLRowe (approved):

DavidLRowe:
I was a little surprised by the double comma in the .model.ts file and also by the zero frequency count for nearly 100 words at the end of the TSV file, but the model seems to compile.

I'm guessing that the words with zero frequency count will be treated as though the count were 1 (which is what happens if the count is missing entirely), but that would be good to test. For future revisions, it would probably be better to include an explicit number.

PR #231, opened 2023-12-13
Added Dictionary to Laz Keyboard

okandale:
A dictionary for my keyboard.


keyman-server:

DavidLRowe:
@okandale Unfortunately there are some problems with the structure of the files for this pull request. You seem to have some extraneous, duplicated files. 

There's also some inconsistencies in naming the files. (The resulting package might still work on your local machine, but it won't work in our build system, which relies on a specific naming scheme.)

It'll probably be easier to close this pull request and create a new one. You'll likely want to use Keyman Developer to create a new lexical model project. (Project menu, New Project, then Wordlist Lexical Model. Use okan_dale as the author, "laz" as the model, and "lzz" as the BCP 47 tag, which will give "okan_dale.lzz.laz" as the base name of the project.

When you submit the new pull request these files should be under the "release" folder (or the "experimental" folder if you prefer). (In the current pull request you have some files in the folder above the "release" folder.)

If you have the possibility of including frequency counts with the words in your wordlist, you might find that the model gives better suggestion.

If you have any questions, or if you'd like me to submit a pull request based on your files, please let me know. Thanks!

PR #233, opened 2023-12-20
Add initial version of okan_dale.lzz.laz_dictionary

okandale:
a dictionary for the Laz keyboard.


keyman-server:

DavidLRowe (approved):

DavidLRowe:
Eventually it would be nice to replace the various places which have a description "...generated from template" with something more informative. You could have something like "Typing dictionary for Laz based on simple word list", for example.

And (as I mentioned on a previous PR), you'll get better suggestions if you include frequency counts with your words. See https://help.keyman.com/developer/16.0/guides/lexical-models/tutorial/step-3 for suggestions on how to do this, assuming that you have some texts in the Laz language that you can use. 

You could even do both: have a frequency-weighted word list (created from some Laz documents) as one file and this existing dictionary (with no frequency information) as a second file. The words in the second file will be given a frequency count of 1. The words in the two files will be combined, thus any words in the dictionary that don't appear in the texts used to make the first list will still show up but at the lowest likelihood of being suggested.

If anything is unclear, please raise an issue.

PR #244, opened 2024-03-05
change: [nrc.en.mtnt] correction-searching now ignores quotes, apostrophes

### LINKED ISSUE #229: bug(nrc.en.mtnt): `we're` not offered as a suggestion for `were`
I think this still needs to remain open for @jahorton to confirm this is resolved:
> With a context of "were", I get 3 suggestions:
"were" (frequency 5309)
"weren't" (frequency 385)
"werent" (frequency 16)
>
>"we're" has a frequency of 927 so I would expect it to appear

On my alpha build of Keyman for Android, I'm still not getting "we're" as a suggestion.

_Originally posted by @darcywong00 in https://github.com/keymanapp/lexical-models/issues/143#issuecomment-1656932828_

Reproduced in Keyman 17.0.219-alpha

jahorton:
Fixes #229.

https://github.com/keymanapp/keyman/pull/10935 includes the updated model in the main repo's "Predictive text - robust testing" KMW test page.

image

image

If we have a perfect match, we run with that:

image

The _real_ test, as someone born and raised in Louisiana:

image

Sadly, the classic `y'all'd've` is not included in this lexical model.  Fortunately, it has the correct `y'all` (freq: 152) listed ahead of the [impostor `ya'll`](https://finance.yahoo.com/news/apos-difference-between-y-apos-185757465.html) (freq: 6).

... it's tempting to go ahead and merge them myself, but that may belong as a separate fix.


mcdurdin (commented):
This needs a version bump in the kps

jahorton:
So, this part does _not_ need a change, right?

https://github.com/keymanapp/lexical-models/blob/9dbc7b7d8a32582e689d45496c7ebff7e2237a8f/release/nrc/nrc.en.mtnt/nrc.en.mtnt.kpj#L11-L17

mcdurdin:
> So, this part does _not_ need a change, right?

Correct, the .kpj metadata is only used by the Keyman Developer IDE.

mcdurdin:
Also HISTORY.md needs updating

DavidLRowe (approved):

PR #241, opened 2024-02-20
[sil.km.ggc] remove space after words

Nnyny:

PR #245, opened 2024-03-06
[sil.km.ggc] bumped version no and added details

MakaraSok:
* bump version number for "[sil.km.ggc] remove space after words #241"


darcywong00 (approved):
lgtm

PR #247, opened 2024-04-04
[nrc.en.mtnt] fix tabs in lexical model source

DavidLRowe:
Two locations had a space instead of a tab in the .tsv file.


darcywong00 (approved):
lgtm

PR #249, opened 2024-04-09
Initial submission of tbm.eo-shaw.esperanto_sxava

tomasbm01:


keyman-server:

tomasbm01:
Thank you. I just fixed the `wordlist.tsv`  and  added a description.

DavidLRowe:
The description change looks good as does the removal of the U+FEFF character. Unfortunately the tab character in the first line of wordlist.tsv has now been replaced by two spaces. I think that will prevent that word from being included. Can you replace those spaces with a tab?

DavidLRowe:
Thanks!

PR #253, opened 2024-05-15
[sil.lazz-latn.laz] change language tag and rename files

Nnyny:

PR #255, opened 2024-05-23
fix: pin @keymanapp/keyman-version to 16.0.138

mcdurdin:
The @keymanapp/lexical-model-compiler package has unpinned versions for @keymanapp/keyman-version and @keymanapp/models-types. This led to a mismatch in module formats from 16.0 to 17.0. This patch tells npm to use the 16.0.138 version of @keymanapp/keyman-version.

I audited the 17.0 packages and found that they all appear to correctly pin the @keymanapp dependency versions in the deployed packages.

Should unblock #253.

Fixes: keymanapp/keyman#11503

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
       "version": "1.0.0",
       "license": "MIT",
       "dependencies": {
+        "@keymanapp/keyman-version": "^16.0.138",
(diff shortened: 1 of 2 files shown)


darcywong00 (approved):
lgtm

PR #260, opened 2024-08-15
teochow model submission

Cathaylaboratory:
here is my teochow model submission

### DIFF
--- a/experimental/ptg/ptg.nan-latn-tw.taigipoj/source/ptg.nan-latn-tw.taigipoj.model.kps
+++ b/experimental/ptg/ptg.nan-latn-tw.taigipoj/source/ptg.nan-latn-tw.taigipoj.model.kps
@@ -20,6 +20,7 @@
     <Copyright URL="">© 2020 PhahTaigi</Copyright>
     <Author URL="">PhahTaigi</Author>
     <Version URL="">1.0</Version>
+    <Description>PhahTaigi POJ Lexical Model</Description>
   </Info>
   <Files>
     <File>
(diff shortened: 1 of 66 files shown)


keyman-server:

DavidLRowe:
Thanks for your submission. 

I think that the project folder name can be just `cathaylab.nan-latn.teochow` (that is, without the `-035` in the language tag), unless there is a different orthography for that region. @mcdurdin Can you confirm?

(And what is region 035?) 

In order to fit in the structure of the repository your project folder (with name Author.LanguageTag.ModelName) needs to be under a folder with just the author name, so `cathaylab.nan-latn.teochow` would be the base folder for the project, and it would be in the folder `cathaylab`, which would in turn be under the `release` folder.

@mcdurdin Do we need to wait on new lexical models while you update the repository to use v17?

Cathaylaboratory:
I think remain the 035 is important because Teochew is different with Taiwanese. 035 means southeast Asia.

Cathaylaboratory:
I have change the name teochow to teochew to recorrect my false spelling. Have you received that?

Sent with [Proton Mail](https://proton.me/) secure email.

mcdurdin:
> @mcdurdin Can you confirm?

I think that the region may be important in this case -- but then the keyboard will need to have that bcp47 listed in its registration also.

> @mcdurdin Do we need to wait on new lexical models while you update the repository to use v17?

No, the repository should be ready for new lexical models.

Cathaylaboratory:
The nan-latn-tw is different with nan-latn-035, I choose 035 to represent teochew because many teochew users lives in Thai, Vietnam, and Canton. I consider there is southeast asia, this model will only match the teochew keyboard. I don't want this model to match nan-latn-tw.

Sent with [Proton Mail](https://proton.me/) secure email.

mcdurdin:
It might be better for compatibility reasons to list each of the regions separately, nan-latn-th, nan-latn-vn etc., rather than using the numeric region code

PR #262, opened 2024-08-16
chore: migrate description field from .model_info to .kps

mcdurdin:
The model info compiler no longer uses the .model_info data, but we never migrated the relevant description field to the .kps file.

Note: the model info compiler should either error on missing Description field or fill in a default value from the Name. (Same for keyboard info compiler.)

Relates-to: keymanapp/keyman#12198
Relates-to: keymanapp/keyman#12202


darcywong00:
> The model info compiler no longer uses the .model_info data,

Is it safe to remove the .model_info files?

darcywong00 (approved):
lgtm

mcdurdin:
> Is it safe to remove the .model_info files?

See #263

Issue #246, opened 2024-03-26
chore: rebuild all lexical models to address missing low-frequency prefix words


mcdurdin:
**Note**: as this is a compiler bug, all models will likely need recompilation so that they may receive the fix's effects.

_Originally posted by @jahorton in https://github.com/keymanapp/keyman/issues/11074_

After 17.0 is released to stable, and the toolchain is deployed to the lexical-models repository, we need to increment the version number on all lexical models in order to get a fixed version of the models to end users.

DavidLRowe:
@mcdurdin or @darcywong00 what is the status of this issue?

darcywong00:
lexical-models compiler updated updated to 17.0.328 in #227.

I'll raise this during A18S9 sprint planning (checking if there's any other compiler issues to get in before rebuilding all models)

jahorton:
My part's done; the compiler fixes needed are in `kmc-model` stable-17.0 already.

darcywong00:
From A18S9 sprint planing, we'll also wait on #263 before rebuilding all the models

PR #265, opened 2024-08-19
fix: fix custom wordbreaker output format for sil.km.gcc

jahorton:
Fixes https://github.com/keymanapp/keyman/issues/12200.

Recent changes in 18.0-alpha require an update to the custom wordbreaker I originally wrote for this model.  We now seek to model whitespaces in the context to ensure we don't accidentally auto-correct away punctuation marks, but those same changes require precise adherence to the wordbreaker output format.  Unfortunately, my original version of it played a bit... fast and loose, returning improper values for blank or duplicated tokens.

### DIFF
--- a/release/sil/sil.km.ggc/source/sil.km.ggc.model.ts
+++ b/release/sil/sil.km.ggc/source/sil.km.ggc.model.ts
@@ -12,16 +12,16 @@ const source: LexicalModelSource = {
   sources: ['wordlist.tsv'],
   wordBreaker: function (str) {
     const whitespaceRegex = /\s|\u200b|\n|\r/;
-    const plainTokens = str.split(whitespaceRegex);
+    const tokens = str.split(whitespaceRegex);
 
-    for(let i=0; i < plainTokens.length; i++) {
-      const token = plainTokens[i];
+    for(let i=0; i < tokens.length; i++) {
+      const token = tokens[i];
       if(token.length == 0) {
-        plainTokens.splice(i, 1);
+        tokens.splice(i, 1);
         i--;
         continue;
       } else if(token.length == 1 && whitespaceRegex.test(token)) {
-        plainTokens.splice(i, 1);
+        tokens.splice(i, 1);
         i--;
         continue;
       }
@@ -48,11 +48,11 @@ const source: LexicalModelSource = {
         const right = token.substring(splitPoint+1);  // Starting past the end of the string => ''
 
         if(left) {
-          plainTokens.splice(i++, 0, left);
+          tokens.splice(i++, 0, left);
         }
-        plainTokens.splice(i++, 1, punct);
+        tokens.splice(i++, 1, punct);
         if(right) {
-          plainTokens.splice(i, 0, right);
+          tokens.splice(i, 0, right);
         }
         // Ensure that the next iteration puts `i` immediately after the punctuation token... even if
         // there was a `right` portion, as it may have extra marks that also need to be spun off.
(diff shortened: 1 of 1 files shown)


release/sil/sil.km.ggc/source/sil.km.ggc.model.ts
```diff
@@ -10,59 +10,71 @@
...
+      const token = plainTokens[i];
+      if(token.length == 0) {
+        plainTokens.splice(i, 1);
+        i--;
+        continue;
+      } else if(token.length == 1 && whitespaceRegex.test(token)) {
+        plainTokens.splice(i, 1);
+        i--;
+        continue;
+      }
```
jahorton: Prevents empty tokens from appearing in the middle of context, especially should excessive whitespaces exist between words.  Something like `the        quick   brown fox` should appear to skip all spaces between tokens at once, not in multiple steps.

release/sil/sil.km.ggc/source/sil.km.ggc.model.ts
```diff
@@ -10,59 +10,71 @@
...
-      // there was a `right` portion, as it may have extra marks that also need to be spun off.
-      i--;
     }
-   }
-   return tokens.map(function(token) {
+
+    let latestIndex = 0;
+    return tokens.map(function(token) {
+      const start = str.indexOf(token, latestIndex);
+      latestIndex = start + token.length;
```
jahorton, line 66: The key part of the fix:  we should track the furthest index reached while iterating over each token and skip past it when determining the original position of the next token to be processed.  This fixes handling for any duplicate-string cases.

release/sil/sil.km.ggc/source/sil.km.ggc.model.ts
```diff
@@ -10,59 +10,71 @@
...
+      const token = tokens[i];
+      if(token.length == 0) {
+        tokens.splice(i, 1);
+        i--;
+        continue;
+      } else if(token.length == 1 && whitespaceRegex.test(token)) {
+        tokens.splice(i, 1);
+        i--;
+        continue;
+      }
```
jahorton, line 27: Prevents empty tokens from appearing in the middle of context, especially should excessive whitespaces exist between words.  Something like `the        quick   brown fox` should appear to skip all spaces between tokens at once, not in multiple steps.

DavidLRowe:
@darcywong00 Would you be able to review this PR? I don't think I'm qualified.

jahorton:
For what it's worth, I've included a copy of the new wordbreaker code in https://github.com/keymanapp/keyman/pull/12229, which is then used in automated tests.  See https://github.com/keymanapp/keyman/blob/ec32e21bf4bd6f7d225126e34deaec7ff728e302/common/models/templates/test/test-tokenization.js#L529-L575.

darcywong00:
> Recent changes in 18.0-alpha require an update to the custom wordbreaker I originally wrote for this model. 

So should these changes get in before we rebuild all the lexical-models in #246?
e.g. Should we bump the gcc-custom-breaker version (in the .kps) here or #246?

release/sil/sil.km.ggc/source/sil.km.ggc.model.ts
darcywong00: Does the custom wordbreaker for sil.km.cnd.model.ts need a similar fix?

https://github.com/keymanapp/lexical-models/blob/a90fec5d323bc30f6fe1783ec604aab95f3d93a2/release/sil/sil.km.cnd/source/sil.km.cnd.model.ts#L13-L25

jahorton:
> > Recent changes in 18.0-alpha require an update to the custom wordbreaker I originally wrote for this model.
> 
> So should these changes get in before we rebuild all the lexical-models in #246? e.g. Should we bump the gcc-custom-breaker version (in the .kps) here or #246?

Either order would likely be fine, but yeah, we'd probably need to bump the version again if done after #246, assuming it includes a version bump.  So... it wouldn't hurt to push this in before the "rebuild all".

mcdurdin (approved):

PR #271, opened 2024-08-23
chore: remove unused .model_info

mcdurdin:
Note: two models had related packages that were missing from .kps.

Fixes: #263
Unblocks: #246

PR #270, opened 2024-08-23
feat: use builder script for models

mcdurdin:
Fixes: #261

PR #272, opened 2024-08-24
chore: add .model_info for burushos.bsk.burushaski

mcdurdin:
This is needed temporarily until #270 merges (and will need removal again in #271).


darcywong00 (approved):
lgtm

PR #273, opened 2024-08-26
chore: Add LICENSE.md files to sample projects

darcywong00:
Follows #271

In preparation for rebuilding all the lexical-models for #246, I noticed sample projects failed to build because they don't contain a LICENSE.md file. (The .kps files already reference the LICENSE.md files)


mcdurdin (approved):
Good catch, LGTM

PR #274, opened 2024-09-09
chore: Rebuild all models with kmc-17.0.329 to address missing low-frequency words

### LINKED ISSUE #246: chore: rebuild all lexical models to address missing low-frequency prefix words

darcywong00:
Fixes #246

* ~Update kmc to latest 17.0.329 stable~ That was split out to #277
* Rebuilds all lexical-models to address missing low-frequency words

Also notes HISTORY.md file of sil.km.ggc for #265


experimental/ptg/ptg.nan-latn-tw.taigipoj/HISTORY.md
```diff
@@ -1,6 +1,10 @@
 PhahTaigi POJ Lexical Model Change History
 ====================
 
+1.1 (2024-09-09)
+----------------
+* Rebuild with 17.0.329 compiler
```
mcdurdin, line 6: I think this needs to say _why_ we are rebuilding with the updated compiler.

mcdurdin (changes requested):

DavidLRowe:
Waiting for @mcdurdin to review and give approval.

mcdurdin:
Can we split the compiler version update into its own PR please? Then this becomes just maintenance, and issues such as the sil.lzz.laz update happening simultaneously to this won't have an impact.

package-lock.json
mcdurdin, line 1: This file shouldn't be in this PR

mcdurdin (approved):
Approved with removal of package-lock.json changes

package-lock.json
darcywong00: argh - I reverted package-lock.json but pushed to my fork instead of here. Will revert on a follow-on

mcdurdin: Don't worry about it. It doesn't do any harm as it stands.

Issue #266, opened 2024-08-21
bug(cathaylab.nan-latn-035.teochew): missing .model_info due to `SkipMetadataFiles` in .kpj


mcdurdin:
This line needs to be removed:

https://github.com/keymanapp/lexical-models/blob/5950a285dd8008af1cf9af196f4012ea1bb43f1d/release/cathaylab/cathaylab.nan-latn-035.teochew/cathaylab.nan-latn-035.teochew.kpj#L7

It caused the [upload to fail](https://build.palaso.org/buildConfiguration/Keyman_Models_BuildAndDeploy/488874?buildTab=log&linesState=3031&logView=flowAware&focusLine=3057) with:

```
22:46:48   Uploading /c/BuildAgent/work/e22cfa4d1a6faf97/models/release/cathaylab/cathaylab.nan-latn-035.teochew/
22:46:48   Failed to locate /c/BuildAgent/work/e22cfa4d1a6faf97/models/release/cathaylab/cathaylab.nan-latn-035.teochew/build/cathaylab.nan-latn-035.teochew.model_info
22:46:48   Aborting with error 1
```

PR #276, opened 2024-09-16
[sil.km.gcc] Fix custom wordbreaker output (version bump to build)

darcywong00:
This follows #265 which fixed the custom wordkbreaker output for sil.km.gcc. We held off on incrementing the version on that PR in anticipation of consolidating it with the repo-wide rebuild of #274.

@mcdurdin requested
> Can we split the compiler version update into its own PR please? Then this becomes just maintenance, and issues such as the sil.lzz.laz update happening simultaneously to this won't have an impact.

So this PR will happen before #266 (which uses 17.0.329), so this build will happen with 17.0.328 stable compiler.

265 266

### DIFF
--- a/release/sil/sil.km.ggc/HISTORY.md
+++ b/release/sil/sil.km.ggc/HISTORY.md
@@ -1,6 +1,10 @@
 Google Crawler Change History
 ====================
 
+1.2 (2024-09-16)
+----------------
+* Fix custom wordkbreaker output format
+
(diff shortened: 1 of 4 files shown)


mcdurdin:
> So this PR will happen before #274 (which uses 17.0.329), so this build will happen with 17.0.328 stable compiler.

Is there any need to push two updates? Can we just use #274 to wrap up both fixes and avoid having users download two updates?

PR #281, opened 2024-10-25
fix[sil.cmo.bw]: follow the improved script of word prediction

Meng-Heng:
fixes: #238


DavidLRowe:
One minor correction and a suggestion for clarity:

LICENSE.md
This is the one file that needs to have the date (or date range). Please use "Copyright © 2021-2024 SIL Global".

HISTORY.md 
The description is somewhat ambiguous. "The script will handle" could be understood to mean that the script is being changed to remove punctuation marks, when the opposite is true. 

Perhaps replace "This script will handle the removal..." with "This script avoids the removal..."

Meng-Heng:
I have applied your suggestion, @DavidLRowe. This PR is ready for a review.

DavidLRowe (approved):

Issue #286, opened 2024-11-04
chore: sil_senegal and sil_cameroon lexical models should be under sil_senegal/ and sil_cameroon/


mcdurdin:
Currently, this is not enforced in the build scripts but we have used some scripts in the past which assume that the prefix will match

So: release/sil/sil_cameroon.mct.mengisa should be release/sil_cameroon/sil_cameroon.mct.mengisa.

And so on.

This should not require a version bump as it is internal to the repo.

I have not checked all the other paths.

PR #289, opened 2024-11-04
[sil_senegal][sil_cameroon] move models to folder that agrees with name

DavidLRowe:
Addresses specific models listed in #286 (but there may be more)


DavidLRowe:
According to @mcdurdin 's comment in that issue, no version increment is required.

DavidLRowe:
A quick check (via script) shows no other such projects (where the author part of the project folder name doesn't match the containing folder).

DavidLRowe:
fixes #286

PR #308, opened 2025-03-08
[jt.nrf-je.spellcheck] add lexical model for Jèrriais

simlrh:
Hello, first time contributor wanting to add a lexical model and a keyboard layout for the Jèrriais language to Keyman. Apologies for anything that I haven't done correctly.

The build system gives a warning that the BCP 47 tag `nrf-je` is not canonical and should be `nrf`. I'm reluctant to use `nrf` because this tag is for "Norman", of which there are at least 3 variants with different orthographies. This lexical model wouldn't be useful to a Guernésiais speaker for example.

Please let me know any changes I need to make.


keyman-server:

DavidLRowe:
Currently you have:
`experimental/jt/jerriais_tech.nrf-je.spellcheck/`
however the first part of the project name (author) needs to match the folder that the project folder is in, so you need to change to either:
`experimental/jerriais_tech/jerriais_tech.nrf-je.spellcheck/` or
`experimental/jt/jt.nrf-je.spellcheck/`
In the latter case you would need to change the filenames as well.

simlrh:
Ready for review

DavidLRowe:
You say "based on the Jèrriais spellcheck file maintained by Geraint Jennings". It would be nice to explicitly say "used with permission" (assuming you do have permission to use this data!). But you can add that at some later date.

DavidLRowe (approved):

PR #310, opened 2025-04-02
chore(chechen_language.ce-latn.chechen): remove https url from email field

mcdurdin:
The author URL field only supports an email address at this time. No version bump required, this impacts only the compiler, picked up with v18.0 beta compiler


darcywong00 (approved):
lgtm

PR #313, opened 2025-05-06
[jra-khmr] update lexical model wordlist

Nnyny:
On behalf of a user.


keyman-server:

Nnyny:
Ready for review.

DavidLRowe:
The description in the jirai.kps file still has the old word count. You may want to remove the word count from the description (which shows up in README.md, readme.htm and the jirai.kps files) so that those files don't need to be changed each time the word list is updated.

The new wordlist.tsv file has no frequency count information (whereas the old file did). Is this intentional? The model will still suggest words, but it won't be able to give priority to the most frequent ones.

jahorton:
> I don't think that that will give the desired result. A word with no count value is assigned a count of 1. Thus all the words in the list will have a count of 1, which would indicate they are all equally likely.
> 
> I suppose you could open the .tsv file in a spreadsheet program (Microsoft Excel or LibreOffice Calc) or in a google sheet, then assign a count of 2588 to the first word, then for the remaining 2587 words use a formula to subtract 1 from the count of the preceeding row (that is, write the formula in the second row, then copy it for all the remaining rows). That would give you counts that decrease from 2588 to 1, so a word earlier in the file would be more likely to be presented as an option. @jahorton Do you think that would work?

It would _kinda_ work.  It's more helpful to know _how much_ more frequent each word is than those after it, not just that it _is_ more frequent, though.

DavidLRowe:
@Nnyny Do you think the user would be able to supply actual word counts for the data?

Nnyny:
I will ask him about this. Thanks!

Nnyny:
I have updated with a new wordlist. Please review.

release/sil/sil.jra-khmr.jarai/source/wordlist.tsv
```diff
@@ -1,1287 +1,2603 @@
...
-ក្នាបតាប	46
-ក្នាល់	3
-ក្មះ	1
-ក្យវ	40
-គុំ	1
-គើញ	1
-ចក់	10
-ចង	1
-ចង់	2
+word	word count
```
jahorton, line 1:
```suggestion
# word	word count
```

DavidLRowe (approved):

PR #328, opened 2025-09-07
GFF Amharic lexical model v1.1.1

dyacob:
This is a small change for the frequency of entry of "ነዉ", which was artificially high due to a common typing error of the appropriate word "ነው".  The consequence has been  "ነዉ" appearing as a top-level suggestion when it shouldn't be for a seldom-used word.

The frequency counts subtracted from  "ነዉ" *could* be added to the intended word "ነው".  I have not done so, but can if advised to do so.  "ነው" is still the most frequently used word, so adding the difference shouldn't change its ranking.


dyacob:
Thank you @darcywong00 , my last commit fixes the version number in the .kps file  Also the copyright year range in the file.

PR #330, opened 2025-10-02
GFF Amharic Lexical Model v2.0

dyacob:
This revision of the GFF Amharic lexical model breaks away from the UniLex resource.  The lexical model is now based on two corpus collections, one for newspapers and one ebooks.  See the [README.md](README.md) for a detailed description.  Some rationale for making this shift:

* UniLex sources are unknown and have unknown provenance.
* UniLex is built from a large corpus of unknown quality and content.
* UniLex content sometimes misidentifies the language.
* The update addresses these problems; all content is now identifiable, traceable, and may be reviewed and revised if issues are found.

**Question:**  In the README.md , does it make sense to have the "Supported Platforms" section?  My understanding is that the dictionaries were device-independent. I would like to remove it if it is superfluous.


DavidLRowe:
I think you are correct that the platform information in the README.md is not needed and can be removed.

DavidLRowe:
Are this ready for review?

DavidLRowe (approved):

PR #351, opened 2026-08-08
[radoslove.isv-latn.wordlist] - Initial creation of Interslavic lexical model

radoslove:
First Interslavic lexical model in the catalogue. Companion to the `isv_latin` keyboard.

39,777 inflected forms, not lemmas. In a language this heavily inflected the distinction decides whether prediction is useful at all: you type `slovami`, not `slovo`.

Derived from the 248,845-form wordlist generated from the medzuslove dictionary database — the same source as the Interslavic HeliBoard dictionary. Trimmed to frequency >= 100 because the full list compiles to a 33 MB trie, past what an iOS keyboard extension can hold. Vocabulary specific to Interslavic itself is kept regardless of frequency: the source corpus scores `medžuslovjansky` at f=26 and `tipkovnica` at 99, so a plain cut-off would drop precisely the words its users type most.

Standard orthography only, matching the keyboard — the suggestion bar never proposes a spelling the standard does not use.

Sources and the generator: https://github.com/radoslove/keyboard-interslavic


keyman-server:

darcywong00:
Thanks for the submission. Do you know if there's any plans to make an Interslavic lexical model using Cyrillic script?

I wonder if we should rename the paths (and references) to `radoslove.isv-latn.wordlist` to distinguish from `isv-cyrl`

release/radoslove/radoslove.isv.wordlist/LICENSE.md
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
+
+© 2026 Radoslove
```
darcywong00: This needs to match the expected standard
```suggestion
Copyright © 2026 Radoslove
```

radoslove: Fixed, thanks.

radoslove:
Agreed on the rename — pushed as `radoslove.isv-latn.wordlist`. You're right that plain `isv` claims the whole language when the model only covers Latin, and Interslavic really is written in both scripts. The companion keyboard is already
`isv_latin`, so the naming lines up. (The licence line came from the `kmc generate` template, so the same slip was in keymanapp/keyboards#4092 — corrected there too.)

On Cyrillic: not in the immediate plans. Cyrillic layouts for Interslavic do exist in my repo, but they've never been audited to the same standard as the Latin one, and I have no way to test a Cyrillic model properly right now. I'd rather submit it as a separate PR once it's actually verified than add an unchecked script to this one.

radoslove:
Ready for review

darcywong00 (approved):
lgtm (looks good to me).
Thanks for the submission!

PR #353, opened 2026-09-16
feat: Add predictive text and autocomplete support for Nuer (nus)

### LINKED ISSUE #345: feat: Add predictive text and autocomplete support for Nuer
### Is your feature request related to a problem? Please describe.

Currently, the Nuer keyboard in Keyman lacks a lexical model. and users do not have access to word suggestions. this is causing manual typing be slow and prone to typos.

### Describe the solution you'd like

I would like to see a predictive text/autocomplete feature added for the Nuer language. This would involve:

Developing or integrating a lexical model (word list) for Nuer.

Ensuring the dictionary accounts for Nuer-specific characters and diacritics.

Enabling the suggestion banner for both mobile and desktop (where supported) versions of the Keyman Nuer keyboard.

### Describe alternatives you've considered

None. A built-in lexical model within Keyman is the most effective way to provide this functionality across all apps.

### Related issues

_No response_

### Keyman apps

- [x] Keyman for Android
- [x] Keyman for iPhone and iPad
- [x] Keyman for Linux
- [x] Keyman for macOS
- [x] Keyman for Windows
- [ ] Keyman Developer
- [ ] KeymanWeb
- [ ] Other - give details at bottom of form

### Keyman version

_No response_

### Operating system

_No response_

### Device

_No response_

### Target application

_No response_

### Browser

_No response_

### Keyboard name

_No response_

### Keyboard version

_No response_

### Language name

_No response_

### Additional context

_No response_

buayism:
- Closes #345
- Integrates optimized book, natural sentence, and biblical wordlists
- Created and compiled by Buay Kun Gach Deng


keyman-server:

release/b/buay_kun_gach_deng.nus.nuer/source/book_wordlist.tsv
```diff
@@ -0,0 +1,4279 @@
...
+ji̱	314
+nɛy	312
+thi̱n	311
+kɛn	308
+bɛ	298
+laa	281
+ɛn	276
+raan	271
+dee	263
+/ci̱	254
```
darcywong00, line 30: Just wanted to confirm if the `/` character is supposed to be here. Generally we don't include punctuation characters in the wordlists.

There may 21 of these in the file.

And 3 more in natural_sentences_wordlist.tsv

buayism: Yes the / character is meant to be there. In Nuer language, when / character appears before a word like "/Cu", it means that word is negative. So "/Cu" and "Cu" are not the same. The one with a slash is negative and the word without a slash is positive

jahorton: While technical, there _is_ a way to resolve this:  https://help.keyman.com/developer/18.0/guides/lexical-models/advanced/unicode-breaker-extension.

Categorizing the `/` as an "ALetter" type instead of its usual punctuation-style handling would let it be part of words.  If care needs to be taken so that this only happens at the start of words and after spaces, that should still be possible, but would require greater fine-tuning.

jahorton: To be clear, I haven't actually tested this within your model directly, but something like the following might be what you're looking for.  

It definitely depends on if you use the `/` character in other ways, like how in English you might see `and/or`, which we interpret as `and`, `/`, `or`.  The example below assumes this pattern exists in your language too, so please update us if that's a bad assumption!

```ts
  wordBreaker: (text) => {
    let customization = {

      /*** Definition of extra word-breaking rules ***/
      rules: [{
        match: (context) => {
          // When encountering [whitespace or start-of-text] + '/' , then a letter,
          // prevent wordbreaks between the '/' and the letter.  Allow them
          // in all other contexts.  (so, a/b would still split into a, /, b)
          if(context.propertyMatch(["WSegSpace", "sot"], ["Negation"], ["ALetter"], null)) {
            return true;
          } else {
            return false;
          }
        },
        breakIfMatch: false
      }],

      /*** Character class overrides for specific characters ***/
      propertyMapping: (char) => {
        if(char == '/') {
            return "Negation";
        } else {
          // The other characters already have useful word-breaking
          // property assignments.
          return null;
        }
      },

      /*** Declares any new, custom character classes to be recognized by the word-breaker ***/
      customProperties: ["Negation"]
    };

    /*** Connects all the pieces together for actual use ***/
    return wordBreakers['default'](text, customization);
  },
```

buayism:
I edited the word breaker block so that / is treated as a negation prefix when it appears at the start of the word or immediately after whitespace. i've built the model successfully and i've confirmed it keeps "/cu" as one token

